PR #49, opened 2016-03-22
Don't depend on Plug or Cowboy

wojtekmach:
Since both Plug & Cowboy are marked as optional dependencies, they shouldn't be required to use. I tried using Swoosh on a project without them and couldn't compile it. This PR fixes it.


stevedomin:
That's a great point @wojtekmach, in the future it probably make sense to move the Plug out of the core codebase, I'll look into that.

barisbalic:
:heart:

PR #53, opened 2016-03-26
Refactor adapter return values

barisbalic:
Most services return useful information, and it makes sense to normalise the return values for each adapter.  This PR would replace all `:ok` returns with `{:ok, map}` where map will usually hold an ID provided by the service.

It's something of a suggestion at the moment.  @stevedomin :eyes: please!

### DIFF
--- a/lib/swoosh/adapters/local.ex
+++ b/lib/swoosh/adapters/local.ex
@@ -22,6 +22,8 @@ defmodule Swoosh.Adapters.Local do
   @behaviour Swoosh.Adapter
 
   def deliver(%Swoosh.Email{} = email, _config) do
-    Swoosh.InMemoryMailbox.push(email)
+    %Swoosh.Email{headers: %{"Message-ID" => id}} = Swoosh.InMemoryMailbox.push(email)
+
+    {:ok, %{id: id}}
   end
 end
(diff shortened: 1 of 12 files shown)


lib/swoosh/adapters/local.ex
```diff
@@ -22,6 +22,11 @@ defmodule Swoosh.Adapters.Local do
   @behaviour Swoosh.Adapter
 
   def deliver(%Swoosh.Email{} = email, _config) do
-    Swoosh.InMemoryMailbox.push(email)
+    id =
```
stevedomin: I'd go for:

```elixir
%Swoosh.Email{headers: %{"Message-ID" => id}} = Swoosh.InMemoryMailbox.push(email)
```

or

```elixir
id = Swoosh.InMemoryMailbox.push(email) |> get_in([:headers, "Message-ID"])
```

stevedomin: Also, `Plug.Swoosh.MailboxPreview` needs more testing to catch that but I believe this change will break it.

stevedomin:
Minor comments, looks good otherwise!

lib/swoosh/adapters/local.ex
barisbalic: I eye-balled the code, but I don't see any reason it would?  The `MailboxPreview` isn't coupled to the delivery in any way...

stevedomin: nvm, might be making that one up

PR #54, opened 2016-03-29
Add support for runtime configuration

### LINKED ISSUE #50: Runtime settings
It seems that at least for the smtp adapter, the env.ex is read once during compilation and those values are compiled in.

Problem being that I'm using System.get_env("SMTP_USERNAME") for example, these values are _not_ available at compile time, but are during runtime (Docker standard behaviour).

This is fine for things like Postgrex and it's database_url setting, or redis configuration, and even mailman - which is what I'm migrating from.

If it's absolutely necessary to keep that behavior for some reason, could there be a deliver method that takes a config (on the mailer and passes that through to the adapter instead of the static configuration)?

stevedomin:
Swoosh.Mailer now supports runtime configuration using
{:system, "ENV_VAR"} tuples. The values will be fetched from the
process environment at runtime.

Fixes #50.

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -25,6 +25,17 @@ defmodule Swoosh.Mailer do
   Most of the configuration that goes into the config is specific to the adapter,
   so check the adapter's documentation for more information.
 
+  Note that the configuration is set into your mailer at compile time. If you
+  need to reference config at runtime you can use a tuple like
+  `{:system, "ENV_VAR"}`.
+
+      config :sample, Sample.Mailer,
+        adapter: Swoosh.Adapters.SMTP,
+        relay: "smtp.sendgrid.net"
+        username: {:system, "SMTP_USERNAME"},
+        password: {:system, "SMTP_PASSWORD"},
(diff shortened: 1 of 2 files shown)


lib/swoosh/mailer.ex
```diff
@@ -53,7 +53,10 @@ defmodule Swoosh.Mailer do
       def deliver(%Swoosh.Email{html_body: nil, text_body: nil}) do
         raise ArgumentError, "expected \"html_body\" or \"text_body\" to be set"
       end
-      def deliver(%Swoosh.Email{} = email), do: @adapter.deliver(email, @config)
+      def deliver(%Swoosh.Email{} = email) do
```
barisbalic: Is this worth multi-lining?

barisbalic:
Makes sense to keep in mind point about reading from ENV on every call, but otherwise :+1:

lib/swoosh/mailer.ex
stevedomin: I multilined it because it felt like this function is the important one and I find it more readable

PR #55, opened 2016-03-29
Support passing config as paramater to deliver/2

### LINKED ISSUE #46: optionally passing config as parameter when sending email
Example: Mailgun. It requires you to use the domain you are sending from in API calls. Some sites use more than one domain.

If you have an app that uses several domains, it would be nice if you could pass the domain as part of the parameters to the sending function.

stevedomin:
For some providers it's useful to be able to override the Mailer
config on a per-email basis. For instance, Mailgun requires you
to specify the domain you are sending from. If you are building
an app that's supporting several domains then you can use that
feature to pass the domain when making the call to deliver/2.

Fixes #46

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -45,6 +45,15 @@ defmodule Swoosh.Mailer do
       %Swoosh.Email{from: {"", "tony@stark.com"}, ...}
       iex> Mailer.deliver(email)
       :ok
+
+  You can also pass an extra config argument to `deliver/2` that will be merged
+  with your Mailer's config:
+
+      # in an IEx console
+      iex> email = %Swoosh.Email{} |> from("tony@stark.com") |> to("steve@rogers.com")
+      %Swoosh.Email{from: {"", "tony@stark.com"}, ...}
+      iex> Mailer.deliver(email, domain: "jarvis.com")
+      :ok
   """
 
   defmacro __using__(opts) do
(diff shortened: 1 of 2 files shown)


stevedomin:
@lau is that what you had in mind?

lib/swoosh/mailer.ex
```diff
@@ -56,19 +56,28 @@ defmodule Swoosh.Mailer do
...
-      def deliver(%Swoosh.Email{subject: nil}), do: raise ArgumentError, "expected \"subject\" to be set"
-      def deliver(%Swoosh.Email{html_body: nil, text_body: nil}) do
+      def deliver(email, config \\ [])
+      def deliver(%Swoosh.Email{from: nil}, _config), do:
+        raise ArgumentError, "expected \"from\" to be set"
+      def deliver(%Swoosh.Email{from: {_name, nil}}, _config), do:
+        raise ArgumentError, "expected \"from\" address to be set"
+      def deliver(%Swoosh.Email{to: nil}, _config), do:
+        raise ArgumentError, "expected \"to\" to be set"
+      def deliver(%Swoosh.Email{to: []}, _config), do:
```
barisbalic: Mixing single/multiline?  Stick to single here if poss.

barisbalic:
Aside from comment :+1:

lib/swoosh/mailer.ex
stevedomin: It's really unreadable, happy to explore other options but this is by far the most readable

lau:
Yeah it looks like it. Would be nice with an example of passing custom run config too. Here: https://github.com/swoosh/swoosh/blob/d4b49ab4f633f72da624e7d96678308ae4709eab/lib/swoosh/mailer.ex#L39

Once this is released I will probably begin looking at switching some production apps I have running from the :mailgun package to Swoosh.

barisbalic:
@lau that's pretty cool.  Keep us posted on how that goes and if anything obvious crops up.

lib/swoosh/mailer.ex
barisbalic: I find it pretty readable, but for now we can crack on.  I'm not happy about mixing the one line/multiline syntax though, `def .. do .. end`

stevedomin:
@lau thanks, good point, I added an example.

If you need any help or needs any bugs fixed you know where to find us

stevedomin:
@lau fyi we've released swoosh 0.2: https://hex.pm/packages/swoosh

lau:
Cool. I'll check it out.

PR #59, opened 2016-04-02
Introduce Email.new/1

stevedomin:
Since `Swoosh.Email` is meant to be an opaque struct it makes sense
to provide a constructor for it.
We allow options to be passed to the function as it can be very
convenient at time (i.e testing).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ defmodule Sample.UserEmail do
   import Swoosh.Email
 
   def welcome(user) do
-    %Swoosh.Email{}
+    new
     |> to({user.name, user.email})
     |> from({"Dr B Banner", "hulk@smash.com"})
     |> subject("Hello, Avengers!")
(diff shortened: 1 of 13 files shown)


stevedomin:
@barisbalic would be good to get your thoughts on this one

barisbalic:
I'm certainly onboard with the idea of providing a 'complete' abstraction, and providing a function that takes a map and returns a populated struct makes sense.  Not sure if it should be a "constructor" however, partly because `new` implies creation of an instance, and partly because of the way it reads.

Wherever `new` is used in the examples looks strange.  I think `Email.new` or even `%Email{}` actually appear clearer to me.  This could just be because I'm not used to seeing it though, do any of the other popular libraries use "constructors"?

stevedomin:
It's definitely used elsewhere:

* [Map.new/0](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/map.ex#L62-L63)
* [MapSet.new/0](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/map_set.ex#L26)
* [Ecto.Multi.new/0](https://github.com/elixir-lang/ecto/blob/master/lib/ecto/multi.ex#L121-L123)

Happy to go with `Email.new` in some of the examples but in the doctests we will have to use either `Swoosh.Email.new/1` or `new` as we can't alias modules there.

I really want to discourage using `%Swoosh.Email{}` going forward.

barisbalic:
Okay sounds good to me, still don't like `new/1` on it's own, but other than that I agree totally :+1:

PR #62, opened 2016-04-08
Allow to assert on email params

stevedomin:
This adds some code to make it possible to use assert_email_sent with certain fields rather than the whole email.


stevedomin:
@barisbalic mind having a look at this one?

barisbalic:
Love it :+1:

PR #67, opened 2016-04-15
Make SMTP adapter add additional headers from Email.headers

nicksanders:
Make SMTP adapter add additional headers from Email.headers

### DIFF
--- a/lib/swoosh/adapters/smtp.ex
+++ b/lib/swoosh/adapters/smtp.ex
@@ -54,6 +54,7 @@ defmodule Swoosh.Adapters.SMTP do
 
   defp prepare_headers(%Email{} = email) do
     []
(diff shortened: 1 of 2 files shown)


stevedomin:
Hi @nicksanders, thanks so much for the PR. That was definitely an oversight.

Before I merge can you:
* Add a test case
* (minor style point in order to be consistent) Use the cons operator `|` to construct the new list, e.g `[headers | Map.to_list(additional_headers)]`

nicksanders:
Hi @stevedomin 

I have added the headers to the test but not sure if the cons operator can be used to concatenate 2 lists?

stevedomin:
Good point.

Thanks a lot for your contribution!

PR #69, opened 2016-04-17
Fixes README.md to match phoenix_swoosh docs

victorsolis:
https://github.com/swoosh/phoenix_swoosh/blob/master/lib/phoenix_swoosh.ex#L2


stevedomin:
Thank you so much!!

PR #70, opened 2016-04-17
Updates README.md `assert_email_sent` example

victorsolis:
I believe this example is clearer than the previous one. The previous example actually fails if you don't construct the exact expected struct.


stevedomin:
That's a great point, can you rebase against master and I'll merge. Thanks so much!!!

victorsolis:
Done, thanks

PR #71, opened 2016-04-18
Allow smtp "MAIL FROM" to be different to "From" header

nicksanders:
For SMTP use the "Sender" header, when it is set, as the "MAIL FROM" instead of the From address.

### DIFF
--- a/lib/swoosh/adapters/smtp.ex
+++ b/lib/swoosh/adapters/smtp.ex
@@ -27,15 +27,19 @@ defmodule Swoosh.Adapters.SMTP do
   @behaviour Swoosh.Adapter
 
   def deliver(%Swoosh.Email{} = email, config) do
-    {_from_name, from_address} = email.from
+    mail_from = mail_from(email)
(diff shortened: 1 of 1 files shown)


stevedomin:
hi @nicksanders 

Thanks for the PR.

While we are at it can we make it so that we fully respect the spec. I think you need to use the `Return-Path` header first if there's one. What do you think of something like:

```elixir
email.headers["Return-Path"] || email.headers["Sender"] || elem(email.from, 1)
```

nicksanders:
Your code looks a lot cleaner than mine but I think the client shouldn't set the return path so we shouldn't use it?

https://tools.ietf.org/html/rfc5321#section-4.4
https://github.com/PHPMailer/PHPMailer/blob/master/class.phpmailer.php line 92

stevedomin:
Yep, you're correct, my apologies.

Can you update the code with the simpler version? And then I'll merge.

stevedomin:
Thanks a lot @nicksanders!!

PR #72, opened 2016-04-21
Add Null adapter

brentonannan:
An adapter that does nothing at all.

Useful for e.g. staging servers where you do not want emails to be sent,
but still want a valid return from `deliver`

The use-case that led to writing this adapter was that I have a long-running staging server, from which I don't want emails to be sent, and where using the `Local` adapter will lead to growing memory usage over the long term, if it's not purged periodically. Rather that setting up this purging, this adapter simply says that it sent an email, but does nothing.

### DIFF
--- a/lib/swoosh/adapters/null.ex
+++ b/lib/swoosh/adapters/null.ex
@@ -1,0 +1,25 @@
+defmodule Swoosh.Adapters.Null do
+  @moduledoc ~S"""
+  An adapter that does nothing at all.
+
+  Useful for e.g. staging servers where you do not want emails to be sent,
+  but still want a valid return from `deliver`
+
+  ## Example
+
+      # config/config.exs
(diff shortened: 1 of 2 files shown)


stevedomin:
Hi @brentonannan 

I had a need for this too at some point, but thought about solving it in a different way, with a Logging adapter. Do you think that would work for you?

brentonannan:
Sorry for the slow reply @stevedomin, national holiday here yesterday =)

A logging adapter would effectively do the same, i.e. not send an email, so if y'all are going to include that in the package, I'd use that in place of this. Just thought it was worth sending a PR since I already wrote this adapter for myself anyways.

stevedomin:
@brentonannan sorry for the time it took us to come around that one, I've put a PR (#78) with a Logger adapter, hopefully that will work for you!

Thanks for sending that PR.

PR #74, opened 2016-04-23
Add dkim support to smtp adapter

nicksanders:
dkim config can be added to config or email.provider_options or both, dkim config in email.provider_options overrides the one in the config.

### DIFF
--- a/lib/swoosh/adapters/sendmail.ex
+++ b/lib/swoosh/adapters/sendmail.ex
@@ -23,8 +23,7 @@ defmodule Swoosh.Adapters.Sendmail do
   @behaviour Swoosh.Adapter
 
   def deliver(%Email{} = email, config) do
-    {type, subtype, headers, parts} = SMTP.prepare_message(email)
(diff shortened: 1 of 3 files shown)


stevedomin:
@nicksanders do you have a use-case for overriding the DKIM options on the per-email basis? If not I'd be quite keen on keeping it simple for now.

nicksanders:
@stevedomin That's fine, I wanted the ability to use different dkim configs depending on the sender domain but I can update the config on deliver. I have also added dkim support to sendmail.

stevedomin:
Thank you so much @nicksanders!!

As a heads up, I might end up extracting the common functionalities from SMTP and Sendmail in a separate module. Haven't figured out where it should live yet.

nicksanders:
Yeah, I had thought that as well, something like RawEmail

barisbalic:
:+1: on the extraction, documenting my opinion even though we've talked about it :D

PR #78, opened 2016-05-14
Add Logger adapter.

stevedomin:
This can be useful when you don't want to send real emails but still
want the Mailer to return successfully.

This was triggered by https://github.com/swoosh/swoosh/pull/72/

### DIFF
--- a/lib/plug/mailbox_preview.ex
+++ b/lib/plug/mailbox_preview.ex
@@ -22,7 +22,8 @@ if Code.ensure_loaded?(Plug) do
     use Plug.Router
     use Plug.ErrorHandler
 
-    alias Swoosh.InMemoryMailbox
+    alias Swoosh.Email.Format
+    alias InMemoryMailbox
 
     require EEx
     EEx.function_from_file :defp, :template, "lib/plug/templates/mailbox_viewer/index.html.eex", [:assigns]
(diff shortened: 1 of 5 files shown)


lib/plug/mailbox_preview.ex
```diff
@@ -22,7 +22,7 @@ if Code.ensure_loaded?(Plug) do
     use Plug.Router
     use Plug.ErrorHandler
 
-    alias Swoosh.InMemoryMailbox
+    alias Swoosh.{Email.Format, InMemoryMailbox}
```
barisbalic: Personally dislike the look of this, but understand it's shorter and likely "idiomatic".

lib/swoosh/email/format.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Swoosh.Email.Format do
```
barisbalic: :+1:

test/swoosh/adapters/logger_test.exs
```diff
@@ -0,0 +1,31 @@
...
+
+  test "deliver/1", %{email: email} do
+    assert capture_log(fn ->
+      {status, _} = LoggerMailer.deliver(email)
+      assert status == :ok
+    end) =~ "New email delivered to steve@rogers.com"
+  end
+
+  test "deliver/1, log full email", %{email: email} do
+    assert capture_log(fn ->
```
barisbalic: Bit messy, but can't think anything nicer.  Maybe define the function then pass it into `capture_log/1` ?

barisbalic:
Love it, depending on how you feel about the comments, get this merged in!

lib/plug/mailbox_preview.ex
stevedomin: Not a huge fan as well, changed it.

test/swoosh/adapters/logger_test.exs
stevedomin: Try both I think this is clearer. That's the style used in the Elixir codebase if that helps convincing.

PR #79, opened 2016-05-14
Move towards storage driver for local adapter

stevedomin:
In the future we want to be able to add other storage driver like disk
and  define a behaviour,
`Swoosh.Adpaters.Local.Storage`, with a couple of callbacks
(all, get, push, etc.) so that people can easily implement their own
driver.

Note that this is a breaking change so will have to communicate about it very clearly in the changelog. (This will only affect people using `Swoosh.InMemoryMailbox` directly).


lib/swoosh/application.ex
```diff
@@ -7,15 +7,15 @@ defmodule Swoosh.Application do
     import Supervisor.Spec, warn: false
 
     children = [
-      worker(Swoosh.InMemoryMailbox, []),
+      worker(Swoosh.Adapters.Local.Storage.Memory, []),
     ]
 
     children =
       if Application.get_env(:swoosh, :serve_mailbox) do
```
barisbalic: Could be extracted to function, potentially two with one overload just returning `children`.  Not strongly against this though.

barisbalic:
👍 Love this too.  Agreed on the changelog, also make it clear that by default `Memory` will "just work", so people don't get put off by the idea of added complexity, 'cos there is none!

lib/swoosh/application.ex
stevedomin: Good point but I'll make the change in a separate PR

Issue #76, opened 2016-05-10
Mailgun Adapter: ArgumentError :erlang.bit_size(nil)


monicao:
Using `swoosh: 0.3.0` and `phoenix_swoosh: 0.1.1`.

I am pretty sure this error is being raised because there is something wrong with my config. It is hard to figure out exactly what is missing from the error message.

```
> MyApp.Emails.forgot_password(%{first_name: "Test", last_name: "User", email: "myemail@gmail.com"}, "test") |> MyApp.Mailer.deliver
** (ArgumentError) argument error
             :erlang.bit_size(nil)
    (swoosh) lib/swoosh/adapters/mailgun.ex:33: Swoosh.Adapters.Mailgun.deliver/2
```

Quick link to the line in question: https://github.com/swoosh/swoosh/blob/v0.3.0/lib/swoosh/adapters/mailgun.ex#L33

Here is the adapter configuration.

```
> Application.get_env(:myapp, MyApp.Mailer)
[adapter: Swoosh.Adapters.Mailgun,
 api_key: "key-a1a2a2s22d2d2d2d2d2d2",
 domain: "mailgun.myapp.com"]
```

And the raw `sys.config` entry just in case this is an issue with strings vs binaries.

```
[
% ... other config stuff here
{myapp, [
  {'Elixir.MyApp.Mailer',[
    {adapter,'Elixir.Swoosh.Adapters.Mailgun'},
    {api_key,<<"key-a1a2a2s22d2d2d2d2d2d2">>},
    {domain,<<"mailgun.myapp.com">>}]}
]},
]
.
```

Any ideas why this error is happening? 

The code works in development where the swoosh config is defined in `config/dev.exs` instead of the `sys.config` file (exrm).

barisbalic:
Hi @monicao thanks for raising this issue.  From what you have said, that sounds correct, although from your example I can't see what's wrong and I can reproduce it.  Would you be up for recreating the issue in a public repo?

monicao:
Sure, I can give that a shot tomorrow.

barisbalic:
That is hugely appreciated @monicao

monicao:
I realized that in order to reproduce the issue I would have to set up the deployment environment (digital ocean droplet, install erlang, etc.). That will be very time consuming and there is no guarantee that it will work.

The email is being sent successfully in development, it's just not sending in production.

I'd be happy to do a quick screen sharing session to debug this on the server I already have. We can set that up over email.

[...]

PR #80, opened 2016-05-15
Raise an error if the mailgun config is missing :domain or :api_key

monicao:
References: https://github.com/swoosh/swoosh/issues/76

Let me know if you have any comments. Thanks.

### DIFF
--- a/lib/swoosh/adapter.ex
+++ b/lib/swoosh/adapter.ex
@@ -10,6 +10,11 @@ defmodule Swoosh.Adapter do
   @typep config :: Keyword.t
 
   @doc """
+  Validates the config passed to the adapter.
+  """
+  @callback validate_config(config) :: {:ok} | {:error, term}
+
+  @doc """
   Delivers an email with the given config.
   """
   @callback deliver(email, config) :: {:ok, term} | {:error, term}
--- a/lib/swoosh/adapters/local.ex
+++ b/lib/swoosh/adapters/local.ex
@@ -21,6 +21,8 @@ defmodule Swoosh.Adapters.Local do
 
   @behaviour Swoosh.Adapter
 
+  def validate_config(_config), do: {:ok}
+
   def deliver(%Swoosh.Email{} = email, config) do
     driver = storage_driver(config)
     %Swoosh.Email{headers: %{"Message-ID" => id}} = driver.push(email)
--- a/lib/swoosh/adapters/logger.ex
+++ b/lib/swoosh/adapters/logger.ex
@@ -26,6 +26,8 @@ defmodule Swoosh.Adapters.Logger do
 
   @behaviour Swoosh.Adapter
 
+  def validate_config(_config), do: {:ok}
+
   def deliver(%Swoosh.Email{} = email, config) do
     rendered_email = render(config[:log_full_email] || false, email)
     Logger.log(config[:level] || :info, rendered_email)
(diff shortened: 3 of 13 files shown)


stevedomin:
I was thinking of adding that back yesterday, so this PR is definitely timely.

Couple of comments:
- I'm not a fan of the `if` statement in deliver/2, it would be nicer to extract the check to a separate function and use it from the Mailer.
- I'm not sure we should `raise` here, maybe we could just return a `{:error, reason}` tuple. @barisbalic what do you think?
- We will need to add this feature to other adapters as well, I can tackle that separately.

Sorry it's quite a lot of comments, let us know if you need any help on this @monicao

monicao:
@stevedomin Thanks for the detailed feedback.

I moved the config validation to `Swoosh.Mailer`. Every adapter can implement a `required_config_keys` function that returns a list of config key names. 

The idea is that each adapter would implement this function and it could be added to the `Swoosh.Adapter` behaviour. Right now I added a function that ignores adapters that do not implement the function. 

Let me know what you think. I am pretty new to elixir so I am still learning the style conventions and best practices.

stevedomin:
@monicao reading your code and having thought more about it I think it could be even simpler (and more flexible). Let's just have a `validate_config` function in each of the adapter (you can leave it as a noop in all the adapters that you don't care about) and add it to the `Swoosh.Adapter` behaviour.

There are a couple of benefits:
- Each adapter is responsible for implementing its own logic, and can do more than checking for required keys.
- It's easier to give more specific error message, eg. "this particular key is missing", rather than "all these keys are required"
- The logic in `Swoosh.Mailer` is much simpler.

I'd still like to hear from @barisbalic (and maybe @lau as well) on the tuple vs raise issue. It would be nice to validate the config when the app starts but since you can pass additional config when calling `deliver/2` we'd have to validate it at each call anyway.

Let me know if you're still keen on working on that, I'm going to be around now so you will have a faster feedback loop.

barisbalic:
Hey all!

I think with regards to the implementation it makes sense for each adapter to house it's own checks.  As for raising.  I am leaning toward returning a tuple of `{:error, "reason"}`, for the reasons @stevedomin stated.

monicao:
@stevedomin @barisbalic Sounds good. I can make the changes.

monicao:
@stevedomin Updated the code and added validations for the Mailgun and SMTP adapters.

stevedomin:
Thanks @monicao, it looks great.

For some reasons CI didn't pick your last commit, can you try rebasing against master, squashing your commits and pushing again?

monicao:
@stevedomin Here it is. Looks like I missed something in the LoggerAdapter, but all the tests are passing now.

barisbalic:
Hi @monicao sorry for taking so long to get back to this!  Could you rebase against master and I'll get this merged in for you.

monicao:
@barisbalic I'm having trouble doing the rebase against master. I think the commits I squashed earlier are back... sigh. If you have time could you give me a hand with the rebase? I added you as a collaborator to the fork.
Otherwise I can create a new fork and make the changes manually.

monicao:
@barisbalic OK, figured out the rebase. Changes from master have been pulled in.

barisbalic:
@monicao thanks for granting me access, if I had had a moment I would have tried to help :(

We really appreciate both your contribution, and your patience, thanks again!

monicao:
Thanks guys! This was fun :)

Issue #56, opened 2016-03-31
Revisit error cases for adapters


barisbalic:
I noticed while doing integration testing that our adapters assume a lot about the response from the provider service.  Specifically if the mailgun domain is invalid or can't be found, the provider returns a 404 with a standard HTML page, which breaks horribly when we try to `Poison.decode!/1` it.

I suspect other providers and errors are better catered for, but we should check, and decide how to handle this in a uniform manner.

PR #81, opened 2016-05-15
Handle 401 response from Mailgun

monicao:
The Mailgun adapter expects a JSON response from the API when the response is a 401 (`:api_key` is invalid). Mailgun returns the string "Forbidden" instead, which causes a exception.

```
test wrong api key (Swoosh.Integration.Adapters.MailgunTest)
     test/integration/adapters/mailgun_test.exs:29
     ** (Poison.SyntaxError) Unexpected token: F
     stacktrace:
       lib/poison/parser.ex:56: Poison.Parser.parse!/2
       lib/poison.ex:83: Poison.decode!/2
       (swoosh) lib/swoosh/adapters/mailgun.ex:37: Swoosh.Adapters.Mailgun.deliver/2
       test/integration/adapters/mailgun_test.exs:39
```

Here is a proposed solution. I didn't actually check that the other status codes also return a string, but I think that's safe to assume that's the case. The docs https://documentation.mailgun.com/api-intro.html#errors are silent.

### DIFF
--- a/lib/swoosh/adapters/mailgun.ex
+++ b/lib/swoosh/adapters/mailgun.ex
@@ -33,6 +33,8 @@ defmodule Swoosh.Adapters.Mailgun do
     case HTTPoison.post(base_url(config) <> "/" <> config[:domain] <> @api_endpoint, params, headers) do
       {:ok, %Response{status_code: code, body: body}} when code >= 200 and code <= 299 ->
         {:ok, %{id: Poison.decode!(body)["id"]}}
+      {:ok, %Response{status_code: code, body: body}} when code == 401 ->
+        {:error, body}
       {:ok, %Response{status_code: code, body: body}} when code >= 400 and code <= 499 ->
         {:error, Poison.decode!(body)}
       {:ok, %Response{status_code: code, body: body}} when code >= 500 and code <= 599 ->
--- a/test/integration/adapters/mailgun_test.exs
+++ b/test/integration/adapters/mailgun_test.exs
@@ -25,4 +25,17 @@ defmodule Swoosh.Integration.Adapters.MailgunTest do
 
     assert {:ok, _response} = Swoosh.Adapters.Mailgun.deliver(email, config)
   end
+
+  test ":error with wrong api key", %{config: config} do
+    config = Keyword.put(config, :api_key, "bad_key")
+
+    email =
+      new
+      |> from({"Swoosh Mailgun", "swoosh@#{config[:domain]}"})
+      |> to("swoosh+to@elixirhq.com")
+      |> subject("Swoosh - Mailgun integration test")
+      |> html_body("<p>This email was sent by the Swoosh library automation testing</p>")
+
+    assert {:error, "Forbidden"} = Swoosh.Adapters.Mailgun.deliver(email, config)
+  end
 end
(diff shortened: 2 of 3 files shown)


monicao:
The CI gets a different error message. I tested this with my Mailgun credentials and got a "Forbidden" response. Not sure why there is a discrepancy here.

stevedomin:
Thanks for the PR @monicao 

Unfortunately we can't do that as the API is sometimes returning JSON errors.

The best way to go about it is to add a separate matching clause just for the 401 status code. Do you want to do that?

Also, the error reported by the CI is in the unit tests, because they expect the error response to be JSON.  The integration tests can only be run for PR or branch belonging to the repo (for security reasons).

monicao:
Thanks @stevedomin, I will make the changes.

monicao:
The CI check failed. Some Mailgun 401 responses must be JSON.

Looking at this comment it seems that the Mailgun API sometimes returns HTML as well. Yay! https://github.com/swoosh/swoosh/issues/56

@stevedomin What do you you think about trying to parse the response as JSON, and if that fails returning the raw response?

stevedomin:
@monicao let's deal with the HTML one in a separate issue/PR.

In order to fix the specs you need to fix the test case for the 401 [here](https://github.com/monicao/swoosh/blob/254a1d00db422b56f41eb3f30d970bf8a6fd11d4/test/swoosh/adapters/mailgun_test.exs#L83).

Let me know if you run into any issue.

monicao:
That test case is failing because it is expecting 401 responses to be HTML. The issue is that sometimes the response is HTML and sometimes it is JSON.

https://github.com/monicao/swoosh/blob/fix_mailgun_response_401/test/integration/adapters/mailgun_test.exs#L29

One way to handle this would be to attempt to parse the JSON and if that doesn't work, return the raw response in `Adapters.Mailgun.deliver`. This approach could be applied to all status codes >= 400.

I wanted to run this by you before making the changes.

stevedomin:
Hi @monicao 

Sorry I'm not sure to understand, I must be missing something. 

The fact that [this spec](https://github.com/monicao/swoosh/blob/254a1d00db422b56f41eb3f30d970bf8a6fd11d4/test/swoosh/adapters/mailgun_test.exs#L83) expects json was a mistake when it was written, based on wrong assumptions.

Now that you've discovered that for 401s Mailgun returns text, and made the appropriate change in the codebase, can we not just change that spec accordingly?

I'm not keen on the try/fail/raw approach, even though Mailgun is doing some weird stuff with their responses we can still know in advance what's going to happen for each status code and create exceptions like this one.

monicao:
It's been a while but for some reason I thought I found a call that returned a 401 and a JSON response. Couldn't reproduce it. Will fix test cases.

monicao:
Looks like the CI did not pick up this commit either. I pulled in the changes from master, squashed the commits and force pushed the branch up. Looks like the CI is still not picking up the commit. I could create a dummy commit and push that up, but that just seems messy. Do you have suggestions?

stevedomin:
@monicao sorry I really don't know what's happening with CI, can you rebase against master and push again please. Hopefully it will work this time. Thanks!

monicao:
@stevedomin Sorry for the delay. The changes have been merged and it looks like all tests passed.

stevedomin:
No worries at all @monicao, thank you for all your contributions :)

PR #83, opened 2016-05-18
update gen_smtp to 0.10.0 - fixes issue with incorrect handling of binary

techgaun:
gen_smtp seems to have the bug which requires to do something like `to_char_list(System.get_env("SMTP_USERNAME"))` right now on swoosh. The [recent release of gen_smtp](https://github.com/Vagabond/gen_smtp/releases/tag/0.10.0) says `Fixes crash in AUTH when username / password is a binary.` I tested this on one of my project and is working fine. thanks!


stevedomin:
Thanks @techgaun!

PR #88, opened 2016-06-19
Use case-insensitive Message-IDs in local storage memory

endersstocker:
This addresses an issue where the mailbox preview was rendered unusable behind an endpoint with lowercase path rewrite (a best practice).

### DIFF
--- a/lib/swoosh/adapters/local/storage/memory.ex
+++ b/lib/swoosh/adapters/local/storage/memory.ex
@@ -34,7 +34,7 @@ defmodule Swoosh.Adapters.Local.Storage.Memory do
       iex> email = new |> from("tony@stark.com")
       %Swoosh.Email{from: {"", "tony@stark.com"}, [...]}
       iex> Memory.push(email)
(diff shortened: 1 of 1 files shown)


lib/swoosh/adapters/local/storage/memory.ex
```diff
@@ -118,7 +118,7 @@ defmodule Swoosh.Adapters.Local.Storage.Memory do
 
   def handle_call({:push, email}, _from, state) do
     id = :crypto.rand_bytes(16) |> Base.encode16
-    email = email |> Swoosh.Email.header("Message-ID", id)
+    email = email |> Swoosh.Email.header("Message-ID", String.downcase(id))
```
stevedomin: Minor comment, can you add the `String.downcase` to the pipeline above?

stevedomin:
Thank you @endersstocker, just one minor comment and I'll merge.

lib/swoosh/adapters/local/storage/memory.ex
endersstocker: 👍 Thanks for catching that.

PR #91, opened 2016-06-27
Enlarge the message preview area.

matthusby:


stevedomin:
Thank you @matthusby!!!

PR #94, opened 2016-07-07
Mailgun provider_option handling custom variables

dighjd12:
I wrote a small method + test that handles custom variables to send with the Mailgun posts. The format expects a map with key the variable name and value the json object. I also tested by sending actual mail and works well. Let me know if there are issues with style or anything!

### DIFF
--- a/lib/swoosh/adapters/mailgun.ex
+++ b/lib/swoosh/adapters/mailgun.ex
@@ -66,7 +66,19 @@ defmodule Swoosh.Adapters.Mailgun do
     |> prepare_cc(email)
     |> prepare_bcc(email)
     |> prepare_reply_to(email)
+    |> prepare_custom_vars(email)
   end
+
+  # example custom_vars
+  # 
+  # %{"my_var" => %{"my_message_id": 123}, 
+  #   "my_other_var" => %{"my_other_id": 1, "stuff": 2}}
+  defp prepare_custom_vars(body, %Email{provider_options: %{custom_vars: my_vars}}) do
+    my_vars 
+    |> Enum.reduce(body, fn({k, v}, body_acc) -> Map.put(body_acc, "v:#{k}", Poison.encode!(v)) end)
+  end   
+  defp prepare_custom_vars(body, _email), do: body
+
 
   defp prepare_from(body, %Email{from: from}), do: Map.put(body, :from, prepare_recipient(from))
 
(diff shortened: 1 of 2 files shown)


lib/swoosh/adapters/mailgun.ex
```diff
@@ -66,8 +66,20 @@ defmodule Swoosh.Adapters.Mailgun do
...
     |> prepare_bcc(email)
     |> prepare_reply_to(email)
+    |> prepare_custom_vars(email)
   end
 
+  # example custom_vars
+  # 
+  # %{"my_var" => Poison.encode!(%{"my_message_id": 123}), 
+  #   "my_other_var" => Poison.encode!(%{"my_other_id": 1, "stuff": 2})}
+  defp prepare_custom_vars(body, %Email{provider_options: custom_vars}) do
```
stevedomin: We probably want to nest this under `custom_vars` or something similar. We may need to pass other provider-specific options that we don't want to end up as custom vars.

lib/swoosh/adapters/mailgun.ex
```diff
@@ -66,8 +66,20 @@ defmodule Swoosh.Adapters.Mailgun do
...
+    custom_vars 
+    |> Enum.reduce(body, fn({k, v}, body_acc) -> Map.put(body_acc, "v:#{k}", v) end)
```
stevedomin: If the custom var data is always JSON is it worth having the encoding step here?

That way you can pass a Map or Keyword list and not have to worry about anything.

lib/swoosh/adapters/mailgun.ex
dighjd12: @stevedomin oops! missed that. I'll address this.

lib/swoosh/adapters/mailgun.ex
dighjd12: @stevedomin I'm thinking you mean we should accept inputs like %{"my_var" => %{"my_message_id": 123}} and do JSON encoding inside our function? I agree that could be better since as far as I know only json values are accepted.

allyraza:
can you squash commits?

dighjd12:
@allyraza done! 💃

princemaple:
updates on this one?

stevedomin:
Thank you @dighjd12!

PR #97, opened 2016-07-25
Drop dependency on httpoison, use hackney directly

michalmuskala:
This removes one dependency from the project, and I don't think it complicates it in any significant way.


stevedomin:
Aside from removing a dependency did you have any other reason for doing that?

Not against that change, just wondering if it's going to make it slightly harder for people to contribute? (Don't really believe it would but better to double check)

michalmuskala:
The only point was to remove the dependency. I don't think httpoison adds anything significant in that situation - there are even 2 lines less in total 😄. I think every additional dependency is a burden, and if it can be easily avoided it should.

stevedomin:
Can't disagree on that, that's fine by me. We can always go back to httpoison if needed.

PR #98, opened 2016-07-28
Make gen_smtp optional

### LINKED ISSUE #90: gen_smtp should be an optional dependency
We should update the docs and README accordingly.

princemaple:
close #90


stevedomin:
Thank you for this PR @princemaple!

Unfortunately it's not that simple, if we make `gen_smtp` optional we need to only compile `Swoosh.Adapters.SMTP` when `:gen_smtp_client` is loaded.

That in itself is quite simple but `Swoosh.Adapters.Sendmail` relies on some functions in `Swoosh.Adapters.SMTP` (this definitely needs to be refactored either way). I'm happy to guide you through the changes if you want to carry on the work on this PR.

princemaple:
> Unfortunately it's not that simple

aha, I merely tried `mix test`. thought it couldn't be that simple.

> if we make gen_smtp optional we need to only compile Swoosh.Adapters.SMTP when :gen_smtp_client is loaded

is that using `Code.ensure_loaded?/1`? something like

```elixir
if Code.ensure_loaded?(:gen_smtp_client) do
  defmodule Swoosh.Adapters.SMTP do
    ....
  end
end
```

> That in itself is quite simple but Swoosh.Adapters.Sendmail relies on some functions in Swoosh.Adapters.SMTP (this definitely needs to be refactored either way). I'm happy to guide you through the changes if you want to carry on the work on this PR.

Sure I can take on that part too, what have you guys thought / decided already?

princemaple:
Regarding the last point, I had a quick scan through the code in `Sendmail` and `SMTP`, looks like I just need to move the mail-processing functions from `SMTP` to a better place like `email.ex` or a new module in `lib/swoosh/email/`. Anything I'm missing?

stevedomin:
> is that using Code.ensure_loaded?/1?

Yes, that's exactly that

> Regarding the last point, I had a quick scan through the code in Sendmail and SMTP, looks like I just need to move the mail-processing functions from SMTP to a better place like email.ex or a new module in lib/swoosh/email/. Anything I'm missing?

That's it as well, however I'm not sure where it should live. `Swoosh.Email` doesn't feel like the right place. I'm not a huge fan of that but in my branch I started moving things to a `SMTP.Helpers` and then used that in both `SMTP` and `Sendmail` adapters. A better way to do it I guess would be to create a protocol for "rendering" the email that all the different adapters could implement.

princemaple:
> A better way to do it I guess would be to create a protocol for "rendering" the email that all the different adapters could implement.

I don't see how a protocol would be applicable here, since the argument we are getting here is always an `%Swoosh.Email{}`. Unless we create an interim adapter specific Email struct for every adapter.

I'll try my best to refactor here.

princemaple:
@stevedomin should be ready now, please take a look when you have a moment. The placement of the extracted code can still be argued about, I'm happy to move it to somewhere else.

P.S. I might have done too much unrelated refactoring.

princemaple:
ping @stevedomin

stevedomin:
Hi @princemaple, sorry for not getting back to you I'm on holidays. Will try to have a thorough look at it later today

princemaple:
:+1:

princemaple:
ping?

lib/swoosh/adapters/local/storage/memory.ex
```diff
@@ -117,7 +117,7 @@ defmodule Swoosh.Adapters.Local.Storage.Memory do
   end
 
   def handle_call({:push, email}, _from, state) do
-    id = :crypto.rand_bytes(16) |> Base.encode16 |> String.downcase
+    id = :crypto.strong_rand_bytes(16) |> Base.encode16 |> String.downcase
```
stevedomin: This change is already in master, do you mind rebasing against it?

lib/swoosh/email/smtp.ex
```diff
@@ -0,0 +1,90 @@
+defmodule Swoosh.Email.SMTP do
```
stevedomin: I'd prefer if this module was closer to the adapters, as it's only used there. Annoyingly I can't really think of a good name for it.

lib/swoosh/email/unify.ex
```diff
@@ -1,13 +1,13 @@
-defmodule Swoosh.EmailHelpers do
+defmodule Swoosh.Email.Unify do
```
stevedomin: Maybe this one should be `Swoosh.Email.Format` and the current `Swoosh.Email.Format` would be renamed to `Swoosh.Email.Render`?

What do you think?

princemaple:
@stevedomin rebased and addressed the comments other than the naming one, I don't have good ideas either.

lib/swoosh/email/unify.ex
```diff
@@ -1,7 +1,7 @@
-defmodule Swoosh.EmailHelpers do
+defmodule Swoosh.Email.Format do
```
stevedomin: I believe you also need to rename the file here?

test/swoosh/email_test.exs
```diff
@@ -238,7 +238,7 @@ defmodule Swoosh.EmailTest do
     assert email == %Email{private: %{phoenix_layout: false}}
   end
 
-  test "format_recipient/1 error messages" do
+  test "unify_recipient/1 error messages" do
```
stevedomin: s/unify_recipient/format_recipient

princemaple:
@stevedomin thanks, my bad, fixed.

lib/swoosh/email/smtp.ex
stevedomin: I'm really not a fan of using the catch-all "Helpers" but I think I'd rather do that for now, so shall we move this module to `Swoosh.Adapters.SMTP.Helpers`, and I'm going to think about a better fitting name

princemaple:
@stevedomin OK changed to `Swoosh.Adapters.SMTP.Helpers`

PR #105, opened 2016-09-15
Fix missing Bcc for SMTP and Sendmail adapters

nicksanders:
Fix missing Bcc for SMTP and Sendmail adapters


barisbalic:
Looks good, many thanks!

PR #106, opened 2016-10-02
Handle templates for postmark

KamilLelonek:
[Postmark](https://postmarkapp.com/) allows using [predefined templates](http://developer.postmarkapp.com/developer-api-templates.html) instead of sending `HTML` or `TEXT` body.

To do that, we need to provide 2 things:

- `TemplateId` - which is the ID of template to use when sending an email defined in Postmark account
- `TemplateModel` - which presents the model to be applied to the specified template to generate `HtmlBody`, `TextBody`, and `Subject`

This PR is supposed to give users a possibility to send email using these templates instead of defining custom templates from the code.

### DIFF
--- a/lib/swoosh/adapters/postmark.ex
+++ b/lib/swoosh/adapters/postmark.ex
@@ -55,10 +55,8 @@ defmodule Swoosh.Adapters.Postmark do
   defp base_url(config),
     do: config[:base_url] || @base_url
 
-  defp api_endpoint([{:template, true} | _]),
-    do: @api_endpoint <> "/withTemplate"
-  defp api_endpoint(_config),
-    do: @api_endpoint
+  defp api_endpoint(config),
+    do: @api_endpoint <> if config[:template], do: "/withTemplate", else: ""
 
   defp prepare_body(email) do
     %{}
--- a/lib/swoosh/adapters/postmark.ex
+++ b/lib/swoosh/adapters/postmark.ex
@@ -64,8 +64,7 @@ defmodule Swoosh.Adapters.Postmark do
     |> prepare_cc(email)
     |> prepare_bcc(email)
     |> prepare_reply_to(email)
-    |> prepare_template_id(email)
-    |> prepare_template_model(email)
+    |> prepare_custom_vars(email)
     |> Poison.encode!()
   end
 
@@ -99,13 +98,21 @@ defmodule Swoosh.Adapters.Postmark do
   defp prepare_html(body, %Email{html_body: nil}),       do: body
   defp prepare_html(body, %Email{html_body: html_body}), do: Map.put(body, "HtmlBody", html_body)
 
-  defp prepare_template_id(body, %{template_id: nil}),         do: body
-  defp prepare_template_id(body, %{template_id: template_id}), do: Map.put(body, "TemplateId", template_id)
-  defp prepare_template_id(body, _email),                      do: body
+  # example custom vars
+  #
+  # %{
+  #   "template_id"    => 123,
+  #   "template_model" => %{"name": 1, "company": 2}
+  # }
+  defp prepare_custom_vars(body, %Email{provider_options: provider_options}),
+    do: Enum.reduce(provider_options, body, &put_in_body/2)
+  defp prepare_custom_vars(body, _email), do: body
 
-  defp prepare_template_model(body, %{template_model: nil}),            do: body
-  defp prepare_template_model(body, %{template_model: template_model}), do: Map.put(body, "TemplateModel", template_model)
-  defp prepare_template_model(body, _email),                            do: body
+  defp put_in_body({key, value}, body_acc) do
+    key = key |> to_string() |> Macro.camelize()
+    val = Poison.encode!(value)
+    Map.put(body_acc, key, val)
+  end
 
   defp make_request(url, headers, params) do
     case :hackney.post(url, headers, params, [:with_body]) do
--- a/lib/swoosh/adapters/postmark.ex
+++ b/lib/swoosh/adapters/postmark.ex
@@ -26,20 +26,11 @@ defmodule Swoosh.Adapters.Postmark do
 
   def deliver(%Email{} = email, config \\ []) do
     headers = prepare_headers(config)
-    params  = email |> prepare_body() |> Poison.encode!()
-    url     = [base_url(config), @api_endpoint]
+    url     = prepare_url(config)
+    params  = prepare_body(email)
 
-    case :hackney.post(url, headers, params, [:with_body]) do
-      {:ok, 200, _headers, body} ->
-        {:ok, %{id: Poison.decode!(body)["MessageID"]}}
-      {:ok, code, _headers, body} when code > 399 ->
-        {:error, {code, Poison.decode!(body)}}
-      {:error, reason} ->
-        {:error, reason}
-    end
+    make_request(url, headers, params)
   end
-
-  defp base_url(config), do: config[:base_url] || @base_url
 
   defp prepare_headers(config) do
     [
@@ -49,6 +40,12 @@ defmodule Swoosh.Adapters.Postmark do
       {"Accept",                  "application/json"}
     ]
   end
+
+  defp prepare_url(config) do
+    [base_url(config), @api_endpoint]
+  end
+
+  defp base_url(config), do: config[:base_url] || @base_url
 
   defp prepare_body(email) do
     %{}
@@ -60,6 +57,7 @@ defmodule Swoosh.Adapters.Postmark do
     |> prepare_cc(email)
     |> prepare_bcc(email)
     |> prepare_reply_to(email)
+    |> Poison.encode!()
   end
 
   defp prepare_from(body, %Email{from: {_name, address}}), do: Map.put(body, "From", address)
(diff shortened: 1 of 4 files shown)


lib/swoosh/adapters/postmark.ex
```diff
@@ -67,26 +75,53 @@ defmodule Swoosh.Adapters.Postmark do
...
+  defp prepare_html(body, %Email{html_body: html_body}), do: Map.put(body, "HtmlBody", html_body)
+
+  # example custom vars
+  #
+  # %{
+  #   "template_id"    => 123,
+  #   "template_model" => %{"name": 1, "company": 2}
+  # }
+  defp prepare_custom_vars(body, %Email{provider_options: provider_options}),
+    do: Enum.reduce(provider_options, body, &put_in_body/2)
```
stevedomin: Feels dangerous to reduce the whole `provider_options` into the body of the request.

Feels like this would be better handled by a specific `prepare_template/2` function that would match exactly on these two keys and insert them into the body.

lib/swoosh/adapters/postmark.ex
```diff
@@ -67,26 +75,53 @@ defmodule Swoosh.Adapters.Postmark do
...
+  defp prepare_custom_vars(body, _email), do: body
+
+  defp put_in_body({key, val}, body_acc) do
+    key = key |> to_string() |> Macro.camelize()
```
stevedomin: See comment above, but in any case we shouldn't use the `Macro.camelize/1` function

test/integration/adapters/postmark_test.exs
```diff
@@ -5,23 +5,42 @@ defmodule Swoosh.Integration.Adapters.PostmarkTest do
 
   @moduletag integration: true
 
-  setup_all do
-    config = [api_key: System.get_env("POSTMARK_API_KEY"), domain: System.get_env("POSTMARK_DOMAIN")]
-    {:ok, config: config}
-  end
+  @config [
```
stevedomin: This changes the behaviour as it will be grabbing the environment variables at compile-time rather than runtime, that can/will cause problem in some environments.

test/integration/adapters/postmark_test.exs
```diff
@@ -5,23 +5,42 @@ defmodule Swoosh.Integration.Adapters.PostmarkTest do
...
+    api_key: System.get_env("POSTMARK_API_KEY"),
+    domain:  System.get_env("POSTMARK_DOMAIN"),
+  ]
+
+  @base_email new
```
stevedomin: See comment about compile-time/runtime above. This would be nicer as a function.

test/swoosh/adapters/postmark_test.exs
```diff
@@ -15,9 +16,11 @@ defmodule Swoosh.Adapters.PostmarkTest do
   """
 
   setup_all do
-    bypass = Bypass.open
-    config = [base_url: "http://localhost:#{bypass.port}",
-              api_key: "jarvis"]
+    bypass = Bypass.open()
```
stevedomin: Style change, should be in a separate PR and across all adapters.

test/swoosh/adapters/postmark_test.exs
```diff
@@ -15,9 +16,11 @@ defmodule Swoosh.Adapters.PostmarkTest do
...
+    config = [
```
stevedomin: Style change, should be in a separate PR and across all adapters.

test/swoosh/adapters/postmark_test.exs
```diff
@@ -31,14 +34,17 @@ defmodule Swoosh.Adapters.PostmarkTest do
...
+        "To"       => "tony.stark@example.com",
+        "From"     => "steve.rogers@example.com",
+        "HtmlBody" => "<h1>Hello</h1>"
+      }
+
       assert body_params == conn.body_params
-      assert "/email" == conn.request_path
-      assert "POST" == conn.method
+      assert "/email"    == conn.request_path
+      assert "POST"      == conn.method
```
stevedomin: All the above: Style change, should be in a separate PR and across all adapters.

test/swoosh/adapters/postmark_test.exs
```diff
@@ -63,51 +69,95 @@ defmodule Swoosh.Adapters.PostmarkTest do
...
+        "ReplyTo"  => "iron.stark@example.com",
+        "TextBody" => "Hello",
+        "HtmlBody" => "<h1>Hello</h1>"
+      }
 
       assert body_params == conn.body_params
-      assert "/email" == conn.request_path
-      assert "POST" == conn.method
+      assert "/email"    == conn.request_path
+      assert "POST"      == conn.method
```
stevedomin: All the above: Style change, should be in a separate PR and across all adapters.

KamilLelonek:
Changes were made accordingly to your remarks.

lib/swoosh/adapters/postmark.ex
KamilLelonek: Isn't it done https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/mailgun.ex#L76-L80 here the same?

lib/swoosh/adapters/postmark.ex
```diff
@@ -8,8 +8,9 @@ defmodule Swoosh.Adapters.Postmark do
 
       # config/config.exs
       config :sample, Sample.Mailer,
-        adapter: Swoosh.Adapters.Postmark,
-        api_key: "my-api-key"
+        adapter:  Swoosh.Adapters.Postmark,
+        api_key:  "my-api-key",
+        template: true # optional
```
KamilLelonek: Not really, we should pick a different URL based on that: https://github.com/KamilLelonek/swoosh/blob/handle-templates-for-postmark/lib/swoosh/adapters/postmark.ex#L57-L61

lib/swoosh/adapters/postmark.ex
stevedomin: Not exactly, in the Mailgun adapter only the `custom_vars` key inside the `provider_options` is reduce on the template.

lib/swoosh/adapters/postmark.ex
stevedomin: What prevents us from checking if `provider_config["template_id"]` is set?

Even if you want to go with the separate `template` config I feel like this should go in the `provider_options` rather than being global.

lib/swoosh/adapters/postmark.ex
KamilLelonek: Anyway, I changed that.

lib/swoosh/adapters/postmark.ex
KamilLelonek: Because constructing endpoint happens earlier - during initialization, not during email delivery. We shouldn't modify endpoint later on but do that when defining an adapter https://github.com/KamilLelonek/swoosh/blob/79e0b7bef1692a9f7ebc92d34a663f16db2d4e1d/lib/swoosh/adapters/postmark.ex#L58-L59

stevedomin: The construction of the endpoint happens when you call `deliver/2`, so it happens every time you send an email.

I really believe this should be attached to the email rather than a config on the adapter.

stevedomin:
Still a couple of style changes that are cluttering the diff, once these are reverted this will be almost ready to be merged

lib/swoosh/adapters/postmark.ex
KamilLelonek: How would you prepare config then? Based on `%Email{}`?

stevedomin: Yes, you can pass the email down to the `api_endpoint/1` function:

```
defp api_endpoint(%Email{provider_options: %{template: true}}), do: @api_endpoint <> "/withTemplate"
defp api_endpoint(_email), do: @api_endpoint
```

lib/swoosh/adapters/postmark.ex
```diff
@@ -26,30 +27,35 @@ defmodule Swoosh.Adapters.Postmark do
 
   def deliver(%Email{} = email, config \\ []) do
     headers = prepare_headers(config)
-    params = email |> prepare_body |> Poison.encode!
```
stevedomin: I would keep this as is.

lib/swoosh/adapters/postmark.ex
```diff
@@ -26,8 +26,9 @@ defmodule Swoosh.Adapters.Postmark do
...
-    url = [base_url(config), @api_endpoint]
+    email   = prepare_body(email)
+    url     = prepare_url(config, email)
```
stevedomin: And use the email (`%Email{}`) here.

lib/swoosh/adapters/postmark.ex
```diff
@@ -39,15 +40,24 @@ defmodule Swoosh.Adapters.Postmark do
...
      {"Accept", "application/json"}]
   end
 
+  defp prepare_url(config, email),
+    do: [base_url(config), api_endpoint(email)]
+
+  defp base_url(config),
+    do: config[:base_url] || @base_url
+
+  defp api_endpoint(%{"TemplateModel" => _, "TemplateId" => _}),
```
stevedomin: Which means you'd get the `%Email{}` struct here.

That way it's consistent with the rest of the file.

KamilLelonek: Done.

PR #108, opened 2016-10-17
Respect from name in the postmark adapter

michalmuskala:
For some reason it wasn't respected. The API documents that it should be: 
> The From and To fields are able to accept name, in the format of `John Doe <email@example.com>`

http://developer.postmarkapp.com/developer-send-api.html#message-format


princemaple:
@stevedomin

stevedomin:
Thanks @michalmuskala, cutting out a new release today

PR #109, opened 2016-10-20
Update README.md re: when to include gen_smtp

princemaple:
```
==> swoosh
Compiling 22 files (.ex)
warning: function :mimemail.encode/2 is undefined (module :mimemail is not available)
  lib/swoosh/adapters/smtp/helpers.ex:17
```

`:mimemail` is defined in `gen_smtp`

Maybe wrap `SMTP.Helpers` and `Swoosh.Adapters.Sendmail` with `Code.ensure_loaded?` as well?

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ end
     end
     ```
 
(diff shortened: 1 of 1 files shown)


stevedomin:
@princemaple good shout, will do that now

PR #111, opened 2016-12-08
Support SendGrid templates and substitutions

ericlathrop:
Howdy! We use SendGrid's templates at work, so I just needed to support a couple of extra fields through `put_provider_option` to make it work. The template data can be passed in like this:

```
|> put_provider_option(:template_id, "some-guid")
|> put_provider_option(:substitutions, %{"-name-" => "Steve Rogers"})
```


stevedomin:
Great, thanks @ericlathrop!

PR #119, opened 2017-03-17
add attachment support

princemaple:
Start adding attachment support with Sparkpost done already

- sparkpost done
- smtp done
- mailgun https://github.com/princemaple/swoosh/commit/5c4c51d7b0bf6758c0bd29d877d676fa91b8817a

cc @stevedomin

PR #121, opened 2017-04-21
Support attachments in Mailgun Adapter

princemaple:
This is a follow up PR to #119 

- [x] Manual test success

### DIFF
--- a/lib/swoosh/adapters/mailgun.ex
+++ b/lib/swoosh/adapters/mailgun.ex
@@ -52,9 +52,8 @@ defmodule Swoosh.Adapters.Mailgun do
 
   defp auth(config), do: Base.encode64("api:#{config[:api_key]}")
 
-  defp content_type(%Email{attachments: nil}), do: "application/x-www-form-urlencoded"
-  defp content_type(%Email{attachments: []}), do: "application/x-www-form-urlencoded"
-  defp content_type(%Email{}), do: "multipart/form-data"
+  defp content_type(%{attachments: []}), do: "application/x-www-form-urlencoded"
+  defp content_type(%{}), do: "multipart/form-data"
 
   defp prepare_body(email) do
     %{}
@@ -66,6 +65,7 @@ defmodule Swoosh.Adapters.Mailgun do
     |> prepare_cc(email)
     |> prepare_bcc(email)
     |> prepare_reply_to(email)
+    |> prepare_attachments(email)
     |> prepare_custom_vars(email)
   end
 
@@ -79,6 +79,10 @@ defmodule Swoosh.Adapters.Mailgun do
   end
   defp prepare_custom_vars(body, _email), do: body
 
+  defp prepare_attachments(body, %{attachments: []}), do: body
+  defp prepare_attachments(body, %{attachments: attachments}) do
+    Map.put(body, :attachment, Enum.map(attachments, &File.read!(&1.path)))
+  end
 
   defp prepare_from(body, %Email{from: from}), do: Map.put(body, :from, prepare_recipient(from))
 
--- a/lib/swoosh/adapters/mailgun.ex
+++ b/lib/swoosh/adapters/mailgun.ex
@@ -29,18 +29,7 @@ defmodule Swoosh.Adapters.Mailgun do
     headers = prepare_headers(email, config)
     url = [base_url(config), "/", config[:domain], @api_endpoint]
 
-    payload = case prepare_body(email) do
-      %{attachments: attachments} = params ->
-        {:multipart,
-         params
-         |> Map.drop([:attachments])
-         |> Enum.map(fn {k, v} -> {to_string(k), v} end)
-         |> Kernel.++(attachments)}
-      no_attachment ->
-        Plug.Conn.Query.encode(no_attachment)
-    end
-
-    case :hackney.post(url, headers, payload, [:with_body]) do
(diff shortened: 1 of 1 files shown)


lib/swoosh/adapters/mailgun.ex
```diff
@@ -79,19 +79,23 @@ defmodule Swoosh.Adapters.Mailgun do
   end
   defp prepare_custom_vars(body, _email), do: body
 
+  defp prepare_attachments(body, %{attachments: []}), do: body
+  defp prepare_attachments(body, %{attachments: attachments}) do
+    Map.put(body, :attachment, Enum.map(attachments, &File.read!(&1.path)))
+  end
 
-  defp prepare_from(body, %Email{from: from}), do: Map.put(body, :from, prepare_recipient(from))
```
stevedomin: Sorry to be annoying @princemaple but if we make that change I'd rather make it across all adapters and in a separate PR for clarity. Does that make sense?

princemaple: Yes, I thought so too. I will leave this out. I have some other stuff across the repo I want to change as well. They can be in different PRs.

lib/swoosh/adapters/mailgun.ex
```diff
@@ -79,19 +79,23 @@ defmodule Swoosh.Adapters.Mailgun do
   end
   defp prepare_custom_vars(body, _email), do: body
 
+  defp prepare_attachments(body, %{attachments: []}), do: body
+  defp prepare_attachments(body, %{attachments: attachments}) do
+    Map.put(body, :attachment, Enum.map(attachments, &File.read!(&1.path)))
```
stevedomin: Reading the docs at https://documentation.mailgun.com/user_manual.html#sending-via-api

It looks like this needs to be done a bit differently. We need as many `attachment` parameter as there are attachments.

This is a `curl` command taken from the docs for sending an email with attachments:

```
curl -s --user 'api:YOUR_API_KEY' \
    https://api.mailgun.net/v3/YOUR_DOMAIN_NAME/messages \
    -F from='Excited User <YOU@YOUR_DOMAIN_NAME>' \
    -F to='foo@example.com' \
    -F cc='bar@example.com' \
    -F bcc='baz@example.com' \
    -F subject='Hello' \
    -F text='Testing some Mailgun awesomness!' \
    --form-string html='<html>HTML version of the body</html>' \
    -F attachment=@files/cartman.jpg \
    -F attachment=@files/cartman.png
```

Am I missing something?

lib/swoosh/adapters/mailgun.ex
stevedomin: 👍

lib/swoosh/adapters/mailgun.ex
princemaple: You are right. I'm in the middle of testing it and fixing the implementation. It's been too long since I had this code, I forgot whether I tested it already. I've added a todo in the PR description.

stevedomin: Not wondering whose fault it is that you haven't looked at that code in a while. Sorry Po...

princemaple: I'll ping you when everything is ready :)

princemaple: Sorry about the immature PR.

stevedomin: Ahah you don't have to be sorry about anything. Tagged you on the Sendgrid PR.

princemaple:
@stevedomin 
As promised, fixed this and tested it manually. Loving my productive Saturday morning!

lib/swoosh/adapters/mailgun.ex
```diff
@@ -27,10 +27,20 @@ defmodule Swoosh.Adapters.Mailgun do
 
   def deliver(%Email{} = email, config \\ []) do
     headers = prepare_headers(email, config)
-    params = email |> prepare_body |> Plug.Conn.Query.encode
     url = [base_url(config), "/", config[:domain], @api_endpoint]
 
-    case :hackney.post(url, headers, params, [:with_body]) do
+    payload = case prepare_body(email) do
```
stevedomin: What do you think of moving this case statement into a function at the end of `prepare_body/1`?

Not too bothered but maybe it would be cleaner?

princemaple: Agreed, will do tomorrow.

princemaple:
@stevedomin refactor is in.

stevedomin:
Thank you so much @princemaple!

ijunaid8989:
Are inline attachments are supported for mailgun adapter?

princemaple:
technically yes. have you tried it?

ijunaid8989:
Yes for me its not working at all.

I tried both ways but only errors are coming.
```elixir
  defp get_attachments(thumbnail) do
    if thumbnail, do: %Swoosh.Attachment{type: :inline, data: thumbnail, filename: "snapshot.jpg"}, else: nil
  end

  defp add_attachment(email, nil), do: email
  defp add_attachment(email, thumbnail) do
    email |> attachment(Swoosh.Attachment.new(thumbnail, filename: "snapshot.jpg", content_type: "image/jpg", type: "inline"))
  end
```

princemaple:
Lets continue the conversation in https://github.com/swoosh/swoosh/issues/285
It's more about in-memory data instead of inline attachments.

PR #125, opened 2017-04-23
Support attachments in Postmark Adapter

princemaple:
I don't use Postmark so can't really test it through.

Docs ref: http://developer.postmarkapp.com/developer-send-api.html#attachments

### DIFF
--- a/lib/swoosh/adapters/postmark.ex
+++ b/lib/swoosh/adapters/postmark.ex
@@ -62,6 +62,7 @@ defmodule Swoosh.Adapters.Postmark do
     |> prepare_text(email)
     |> prepare_cc(email)
     |> prepare_bcc(email)
+    |> prepare_attachments(email)
     |> prepare_reply_to(email)
     |> prepare_template(email)
   end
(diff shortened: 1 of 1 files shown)


lib/swoosh/adapters/postmark.ex
```diff
@@ -76,6 +77,15 @@ defmodule Swoosh.Adapters.Postmark do
   defp prepare_bcc(body, %Email{bcc: []}), do: body
   defp prepare_bcc(body, %Email{bcc: bcc}), do: Map.put(body, "Bcc", prepare_recipients(bcc))
 
+  defp prepare_attachments(body, %Email{attachments: []}), do: body
+  defp prepare_attachments(body, %Email{attachments: attachments}) do
+    Map.put(body, "Attachments", Enum.map(attachments, &%{
+      "Name" => &1.filename,
+      "ContentType" => &1.content_type,
+      "Content" => &1 |> File.read! |> Base.encode64
```
stevedomin: Shouldn't this be `&1.path |> File.read! |> Base.encode64`?

princemaple: Yep

PR #126, opened 2017-04-23
Support attachments in Mandrill adapter

princemaple:
Docs ref: https://mandrillapp.com/api/docs/messages.JSON.html#method=send suggests

I don't use mandrill either, can't test...

### DIFF
--- a/lib/swoosh/adapters/mandrill.ex
+++ b/lib/swoosh/adapters/mandrill.ex
@@ -96,7 +96,7 @@ defmodule Swoosh.Adapters.Mandrill do
     Map.put(body, "attachments", Enum.map(attachments, &%{
       "name" => &1.filename,
       "type" => &1.content_type,
-      "content" => &1 |> File.read! |> Base.encode64
+      "content" => &1.path |> File.read! |> Base.encode64
     }))
   end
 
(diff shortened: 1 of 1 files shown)


stevedomin:
Trying now

lib/swoosh/adapters/mandrill.ex
```diff
@@ -90,11 +91,20 @@ defmodule Swoosh.Adapters.Mandrill do
   defp prepare_bcc(body, %Email{bcc: []}), do: body
   defp prepare_bcc(body, %Email{bcc: bcc}), do: prepare_recipients(body, bcc, "bcc")
 
+  defp prepare_attachments(body, %Email{attachments: []}), do: body
+  defp prepare_attachments(body, %Email{attachments: attachments}) do
+    Map.put(body, "attachments", Enum.map(attachments, &%{
+      "name" => &1.filename,
+      "type" => &1.content_type,
+      "content" => &1 |> File.read! |> Base.encode64
```
stevedomin: Shouldn't this be `&1.path |> File.read! |> Base.encode64`?

stevedomin:
I just realised I can't try the Mandrill adapter because Mandrill has no free plan :/

lib/swoosh/adapters/mandrill.ex
princemaple: My bad.

Issue #129, opened 2017-05-19
Can't customize SMTP port used by SMTP Adapter


doughsay:
For local development purposes I'm trying to consolidate all email delivery to a locally running [Mailhog](https://github.com/mailhog/MailHog) instance.  By default it runs its SMTP server on port 1025.  The Swoosh SMTP adapter doesn't seem to be able to customize the port used and always assumes 25.

I looked into how the config is passed to gen_smtp, and it LOOKS like all I need to do is add `port: 1025` to my config and it should theoretically work.  But I haven't been able to get it to work.

Any ideas?

princemaple:
> I looked into how the config is passed to gen_smtp, and it LOOKS like all I need to do is add port: 1025 to my config and it should theoretically work.

That's what I thought, too. Do you have any error logs?

valentinewallace:
Adding the port to my config worked for me, I'm able to get emails. 

Here's my config:
`config :coherence, YourApp.Coherence.Mailer,
  adapter: Swoosh.Adapters.SMTP,
  relay: "smtp.gmail.com",
  port: 587,
  username: [your-username],
  password: [your-password],
  tls: :always,
  auth: :always
`

stevedomin:
@doughsay did you manage to get it working?

stevedomin:
@doughsay if this is still an issue please reopen!

Sathras:
i seem to have the same issue - at least i have a custom port.
using phoenix 1.3 and coherence.
It says email send, no error log; but the email never gets sent.

princemaple:
if you guys want to try out #162

PR #133, opened 2017-05-30
Find Adapter and Mailer config at runtime

cjab:
The macro that sets `@adapter` and `@config` on `Mailer` is run at compile time. This causes some confusing side effects. If the config is changed but the Mailer that uses the macro is not the new config will not be used.

This looks up configuration at runtime rather than compile time.

I think this may also be the cause of #129, although I'm not sure.

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -60,16 +60,18 @@ defmodule Swoosh.Mailer do
 
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
-      {otp_app, adapter, config} = Swoosh.Mailer.parse_config(__MODULE__, opts)
+      Swoosh.Mailer.parse_config(__MODULE__, opts)
+      @opts opts
 
-      @adapter adapter
-      @config config
-
-      def __adapter__, do: @adapter
+      def __adapter__ do
+        {_, adapter, _} = Swoosh.Mailer.parse_config(__MODULE__, @opts)
+        adapter
+      end
 
       def deliver(email, config \\ [])
       def deliver(email, config) do
-        Swoosh.Mailer.deliver(@adapter, email, Keyword.merge(@config, config))
+        {_otp_app, adapter, base_config} = Swoosh.Mailer.parse_config(__MODULE__, @opts)
+        Swoosh.Mailer.deliver(adapter, email, Keyword.merge(base_config, config))
       end
 
       def deliver!(email, config \\ [])


princemaple:
> If the config is changed but the Mailer that uses the macro is not the new config will not be used.

This is not the case for me, the files correctly recompile every time I change my config, with no changes made to the mailer.

Could you reproduce the scenario and provide more information? e.g. the detailed reproduce procedure / your setup

cjab:
Thanks for taking a look @princemaple. I've created a repo containing the simplest possible project I could reproduce this with. It only seems to be a problem for umbrella projects, non-umbrella mix projects recompile correctly when the config is changed.

The more I look at it the more I question if it's a problem with mix or maybe the default mix configuration for umbrella projects. That said, here is the repo: https://github.com/cjab/test_umbrella and the steps to reproduce (also in the README):

1) Note `apps/test_swoosh/config/config.exs` is configured to use `Swoosh.Adapters.Test`.

2) Run `mix deps.get && iex -S mix`

3) Run
   ```
   iex> TestSwoosh.UserEmail.welcome() |> TestSwoosh.Mailer.deliver()
   ```
   `{:ok, %{}}` is returned meaning that `Swoosh.Adapters.Test` is being used.

4) Kill iex

5) Change configuration in `apps/test_swoosh/config/config.exs` to
   ```
   config :test_swoosh, TestSwoosh.Mailer,
     adapter: Swoosh.Adapters.Local
   ```

6) Run `iex -S mix`. This should recompile, but it doesn't. Maybe this is a problem with mix?

7) Run
   ```
   iex> TestSwoosh.UserEmail.welcome() |> TestSwoosh.Mailer.deliver()
   ```
   `{:ok, %{}}` is returned meaning that `Swoosh.Adapters.Test` is _still_ being used.

princemaple:
Hi @cjab 

Thank you very much for the detailed reproduce steps and the setup. They are very, very helpful.
Yes, it indeed looks like it's having problems in umbrella apps.
And neither `mix deps.compile` nor `mix compile` do anything useful.

I'd like to grab some attention from @stevedomin , see how he would like to approach this.
Meanwhile, I'll look into your implementation and lookup what's the norms with umbrella apps.

Thanks again.

Po

stevedomin:
@cjab @princemaple sorry for the delay, haven't had time to dive into this issue yet

princemaple:
A relatively simple workaround is to lift the config to root config. Not sure how undesirable this would be :)

wmnnd:
I am reading environment variables in a custom config module that runs on Application start and this pull request does the trick for me.

Are you considering merging it into the official version, @stevedomin?

stevedomin:
@wmnnd hi! yes, it will in the next release.

princemaple:
This will be implemented in #146

PR #134, opened 2017-05-30
Support for mail headers in Mailgun adapter

bhserna:
This pull request adds support for mail headers in the Mailgun adapter, sending the headers added with the function `Email.header/3` according with the Mailgun documentation (https://documentation.mailgun.com/api-sending.html#sending)

> h:X-My-Header h: prefix followed by an arbitrary value allows to append a custom MIME header to the message (X-My-Header in this case). For example, h:Reply-To to specify Reply-To address.


princemaple:
Thanks @bhserna !

bhserna:
@princemaple Thank you for fix it and merge it!

PR #135, opened 2017-05-30
Allow adding attachments in Email.new

princemaple:
I missed this one!

@stevedomin


stevedomin:
@princemaple please rebase against master and this is good to go!

stevedomin:
good catch btw!

PR #142, opened 2017-07-04
Fix typo bug in Sendgrid adapter

stevedomin:


lib/swoosh/adapters/sendgrid.ex
```diff
@@ -103,7 +103,7 @@ defmodule Swoosh.Adapters.Sendgrid do
     Map.put(body, :content, content)
   end
   defp prepare_content(body, %{html_body: html}), do: Map.put(body, :content, [%{type: "text/html", value: html}])
-  defp prepare_content(body, %{text_body: text}), do: Map.put(body, :content, [%{type: "text/plain", type: text}])
+  defp prepare_content(body, %{text_body: text}), do: Map.put(body, :content, [%{type: "text/plain", value: text}])
```
princemaple: I might just throw away these 2 clauses and go with
``` elixir
   defp prepare_content(body, email) do
     content = cond do
       email.html && email.text -> [%{type: "text/plain", value: text}, %{type: "text/html", value: html}]
       email.html -> [%{type: "text/html", value: html}]
       email.text -> [%{type: "text/plain", value: text}]
     end
     Map.put(body, :content, content)
   end
```

stevedomin: Much better. Will change the code.

stevedomin: Actually I think the code you have there already works. I just need to remove these two additional functions

princemaple: Right, that's struct, so the fields are always there. 👍

PR #143, opened 2017-07-04
[WIP] Uniformisation of adapter errors

stevedomin:
One of Swoosh's goal is to provide a unified layer for communicating with transactional email service provider.

We have managed to do that pretty successfully for happy paths but if the delivery fails for any reason then each adapter uses different return type and we lose that unification. It makes it slightly harder to change provider for instance.

This PR is an attempt to solve this issue.


lib/swoosh/adapter.ex
```diff
@@ -22,6 +24,21 @@ defmodule Swoosh.Adapter do
         expected #{inspect key} to be set, got: #{inspect config}
         """
       end
+
+      defp code_to_reason(400), do: :bad_request
+      defp code_to_reason(401), do: :unauthorized
```
yordis: @stevedomin wouldn't be better to remove all the functions out of the macro and do some `import` or something inside?

stevedomin: Hi @yordis, sorry for the slow response, been travelling.

Definitely, I still need to make a lot of improvement to that PR.

PR #146, opened 2017-07-27
Full runtime config support

### LINKED ISSUE #84: Support Conform Style Configuration
The macro reads the configuration and sets it to the module attribute `@config`. Swoosh supports transforming environment variables, but that's only one way of deploying.

[bitwalker/conform](https://github.com/bitwalker/conform) is another model that overwrites the sys.config based on a configuration file deployed to the server. If the config was read at runtime, instead of compile time then the conform style of deployments could be supported as well.

princemaple:
Configs are loaded at runtime.
Module configs are kept, still having higher priority.

Configs now are read in the following order, the later ones take priority:

1. mix configs
```elixir
config :my_app, MyMailer, # the configs
```

2. module configs
```elixir
defmodule MyMailer do
  use Swoosh.Mailer, otp_app: :my_app, # other configs
end
```

3. function configs
```elixir
MyMailer.deliver(email, the_custom_configs)
```

4. system env
All the `{:system, "ENV_VAR"}` are finally loaded

close #84

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -66,14 +66,12 @@ defmodule Swoosh.Mailer do
 
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
-      {otp_app, module_config} = Swoosh.Mailer.parse_static_config(__MODULE__, opts)
-
-      @otp_app otp_app
-      @module_config module_config
+      @otp_app Keyword.fetch!(opts, :otp_app)
+      @mailer_config opts
 
       def deliver(email, config \\ [])
       def deliver(email, config) do
-        Swoosh.Mailer.deliver(email, {@otp_app, __MODULE__, @module_config, config})
+        Swoosh.Mailer.deliver(email, {@otp_app, __MODULE__, @mailer_config, config})
       end
 
       def deliver!(email, config \\ [])
@@ -94,13 +92,8 @@ defmodule Swoosh.Mailer do
       when address in ["", nil] do
     {:error, :from_not_set}
   end
-  def deliver(%Swoosh.Email{} = email, {otp_app, mailer, module_config, config}) do
-    config =
-      Application.get_env(otp_app, mailer, [])
-      |> Keyword.merge(module_config)
-      |> Keyword.merge(config)
-      |> Swoosh.Mailer.parse_system_env
-
+  def deliver(%Swoosh.Email{} = email, unprocessed_config) do
+    config = parse_runtime_config(unprocessed_config)
     adapter = Keyword.fetch!(config, :adapter)
 
     :ok = adapter.validate_config(config)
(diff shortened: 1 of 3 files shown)


lib/swoosh/mailer.ex
```diff
@@ -82,14 +87,21 @@ defmodule Swoosh.Mailer do
...
   end
-  def deliver(_adapter, %Swoosh.Email{from: {_name, address}}, _config) when address in ["", nil] do
+  def deliver(%Swoosh.Email{from: {_name, address}}, _config)
+      when address in ["", nil] do
     {:error, :from_not_set}
   end
-  def deliver(adapter, %Swoosh.Email{} = email, config) do
-    config = Swoosh.Mailer.parse_runtime_config(config)
+  def deliver(%Swoosh.Email{} = email, {otp_app, mailer, module_config, config}) do
+    config =
```
stevedomin: Probably worth extracting this into its own function, `parse_runtime_config` maybe?

lib/swoosh/mailer.ex
```diff
@@ -82,14 +87,21 @@ defmodule Swoosh.Mailer do
...
     {:error, :from_not_set}
   end
-  def deliver(_adapter, %Swoosh.Email{from: {_name, address}}, _config) when address in ["", nil] do
+  def deliver(%Swoosh.Email{from: {_name, address}}, _config)
+      when address in ["", nil] do
     {:error, :from_not_set}
   end
-  def deliver(adapter, %Swoosh.Email{} = email, config) do
-    config = Swoosh.Mailer.parse_runtime_config(config)
+  def deliver(%Swoosh.Email{} = email, {otp_app, mailer, module_config, config}) do
```
stevedomin: Let's make it very explicit that `module_config` is a compile-time config by renaming it to something like `compiled_mailer_config` or `compiled_module_config`.

lib/swoosh/mailer.ex
```diff
@@ -98,26 +110,26 @@ defmodule Swoosh.Mailer do
   @doc """
   Parses the OTP configuration at compile time.
   """
-  def parse_config(mailer, opts) do
-    otp_app = Keyword.fetch!(opts, :otp_app)
+  def parse_static_config(mailer, module_config) do
+    otp_app = Keyword.fetch!(module_config, :otp_app)
     config = Application.get_env(otp_app, mailer, [])
-    adapter = opts[:adapter] || config[:adapter]
```
stevedomin: I don't think we should do this here anymore (getting the adapter module and raising an error if it's not set), it's something that should happen at runtime, in `deliver/2` or whatever function we use for parsing/merging config.

stevedomin: In fact we're doing half of this line 104. I'd extract that and the check/raise in a separate function as well

princemaple: removed and added a test to ensure raise on KeyError with `~r/:adapter not found/`
I was trying to catch this on compile time still, but given now users can pass adapter at runtime, I guess it's no longer necessary.

princemaple:
@stevedomin please take another look

- extracted the config parsing into a function, calling it `parse_config` because it's taking both compile time and run time into account
- the only thing being checked at compile time is that `otp_app` is set, `adapter` is no longer checked
- adapter is `fetch!`'ed at run time

stevedomin:
Looks good!

stevedomin:
@princemaple pretty interesting article about config in Elixir libraries from @michalmuskala: http://michal.muskala.eu/2017/07/30/configuring-elixir-libraries.html

PR #147, opened 2017-07-27
Use Elixir 1.5 final

princemaple:


princemaple:
Thought it would fail. All good.

stevedomin:
Thanks @princemaple!

PR #149, opened 2017-08-05
Upgrade bypass

princemaple:
@stevedomin 
looks like the bypass server is now killed after every test run. not sure if it's a change in Elixir 1.5 or Bypass.


stevedomin:
Nice @princemaple!

PR #150, opened 2017-08-16
Quote name portion of recipient for Mailgun

tmjoen:
If the name portion of a recipient contains a comma, weird things will happen. With multiple recipients, some of them will be chopped off without any warnings or errors.

This changes the formatting from

`Acme Inc, headquarters <hq@acme.com>`

to

`"Acme Inc, headquarters" <hq@acme.com>`

Ref: https://stackoverflow.com/questions/35370571/how-to-add-a-comma-to-a-mailgun-from-field-for-outbound-emails

Ref: http://www.rfc-editor.org/rfc/rfc2822.txt

> Note that the display names for Joe Q. Public and Giant; "Big" Box needed to be enclosed in double-quotes because the former contains the period and the latter contains both semicolon and double-quote characters (the double-quote characters appearing as quoted-pair construct).

This bit me HARD, so I hope you'll consider this PR.

Thanks!

### DIFF
--- a/lib/swoosh/adapters/mailgun.ex
+++ b/lib/swoosh/adapters/mailgun.ex
@@ -116,7 +116,7 @@ defmodule Swoosh.Adapters.Mailgun do
   end
 
   defp prepare_recipient({"", address}), do: address
-  defp prepare_recipient({name, address}), do: "#{name} <#{address}>"
+  defp prepare_recipient({name, address}), do: ~s("#{name}" <#{address}>)
 
   defp prepare_subject(body, %{subject: subject}), do: Map.put(body, :subject, subject)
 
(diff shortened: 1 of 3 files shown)


princemaple:
Hi @tmjoen , thanks for the first time contribution! The test failure seems like an easy fix. If you could take care of that, I'll merge this PR right in. If not, I'll fix it tomorrow (Australia time).

@stevedomin this will need a patch release.

stevedomin:
Hello @tmjoen,

Thanks for the report and PR. This definitely needs to be fixed.

Your PR looks great, I've just one minor thing to ask. This bug was fixed in the [Postmark adapter](https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/postmark.ex#L99) and I'd like to use a sigil there as well. Can you make the change?

Thanks

stevedomin:
Ah, didn't see @princemaple reply :)

I can take care of the test failure don't worry!

tmjoen:
Thanks for the fast replies!

I pushed some fixes

stevedomin:
Thanks @tmjoen! It's perfect.

We'll push a minor version out now.

stevedomin:
@tmjoen version 0.9.1 is out: https://hex.pm/packages/swoosh

PR #156, opened 2017-10-13
Avoid warnings in sample code

denispeplin:


princemaple:
Thanks @denispeplin !

PR #158, opened 2017-10-15
Add typespec to injected code

### LINKED ISSUE #157: add typespec to Swoosh.Mailer
Hey I notice that the type specs are missing so when I generate my documentation I can't see anything in the module that `use` this module.

princemaple:
closes #157


stevedomin:
Thank @princemaple!!!

PR #162, opened 2017-10-18
Enforce config values having correct types before passing to gen_smtp

### LINKED ISSUE #161: Can't customize SMTP port
I pretty much have the same use case and issue as #129.

My config and the error I'm seeing are below.

My config:

```elixir
config :identity_context, IdentityContext.Mailer,
  adapter: Swoosh.Adapters.SMTP,
  password: { :system, "SMTP_PASSWORD" },
  port: { :system, "SMTP_PORT" },
  relay: { :system, "SMTP_HOST" },
  username: { :system, "SMTP_USERNAME" }
```

My error:

```
{:error,
 {:retries_exceeded,
  {:network_failure, 'app.mailhog', {:error, :econnrefused}}}}
```

I had a feeling it might be because the port is coming in from the config as a string and `gen_smtp` might not handle this properly. When I switched to a hardcoded integer in my config I got the following error:

```
** (FunctionClauseError) no function clause matching in :gen_smtp_client.to_string/1    
    
    The following arguments were given to :gen_smtp_client.to_string/1:
    
        # 1
        nil
    
    (gen_smtp) /usr/src/app/deps/gen_smtp/src/gen_smtp_client.erl:338: :gen_smtp_client.to_string/1
    (gen_smtp) /usr/src/app/deps/gen_smtp/src/gen_smtp_client.erl:320: :gen_smtp_client.try_AUTH/3
    (gen_smtp) /usr/src/app/deps/gen_smtp/src/gen_smtp_client.erl:204: :gen_smtp_client.do_smtp_session/3
    (gen_smtp) /usr/src/app/deps/gen_smtp/src/gen_smtp_client.erl:143: :gen_smtp_client.try_smtp_sessions/4
    (swoosh) lib/swoosh/adapters/smtp.ex:43: Swoosh.Adapters.SMTP.deliver/2
    (identity_context) lib/identity_context/registration/command.ex:31: IdentityContext.Registration.Command.send_welcome_email/1
```

princemaple:
Fixes #161

### DIFF
--- a/lib/swoosh/adapters/smtp.ex
+++ b/lib/swoosh/adapters/smtp.ex
@@ -55,9 +55,9 @@ if Code.ensure_loaded?(:gen_smtp_client) do
     @config_transformations %{
       port: &String.to_integer/1,
       retries: &String.to_integer/1,
-      ssl: {&String.to_existing_atom/1, [true, false]},
-      tls: {&String.to_existing_atom/1, [:always, :never, :if_available]},
-      auth: {&String.to_existing_atom/1, [:always, :if_available]}
+      ssl: {&String.to_atom/1, [true, false]},
+      tls: {&String.to_atom/1, [:always, :never, :if_available]},
+      auth: {&String.to_atom/1, [:always, :if_available]}
     }
     @config_keys Map.keys(@config_transformations)
 
--- a/lib/swoosh/adapters/smtp.ex
+++ b/lib/swoosh/adapters/smtp.ex
@@ -20,6 +20,7 @@ if Code.ensure_loaded?(:gen_smtp_client) do
           password: "ilovepepperpotts",
           tls: :always,
           auth: :always,
+          port: 1025,
           dkim: [
             s: "default", d: "domain.com",
             private_key: {:pem_plain, File.read!("priv/keys/domain.private")}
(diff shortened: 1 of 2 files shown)


lib/swoosh/adapters/smtp.ex
```diff
@@ -40,13 +41,51 @@ if Code.ensure_loaded?(:gen_smtp_client) do
...
         receipt when is_binary(receipt) -> {:ok, receipt}
         {:error, type, message} -> {:error, {type, message}}
         {:error, reason} -> {:error, reason}
       end
     end
 
+    @config_transformations %{
+      port: &String.to_integer/1,
+      retries: &String.to_integer/1,
+      ssl: {&String.to_existing_atom/1, [true, false]},
```
stevedomin: why are you using `to_existing_atom/1` here? (just curious)

stevedomin:
@princemaple looks great, just a minor suggestion if you find the time.

princemaple:
Yep, was just trying to see if it works first.

lib/swoosh/adapters/smtp.ex
princemaple: So the user cannot misspell the config and create a new one.
While it doesn't really matter, as it's caught immediately, I always use `to_existing_atom` if `to_atom` is not absolutely needed.

stevedomin: 👍

princemaple:
OK this seems to work well. I tested it by setting up a new test project, traced through with IEx.break! and pry. I'll add some tests and merge.

lib/swoosh/adapters/smtp.ex
princemaple: I changed this to `to_atom` in the end:

- this is not from user input, should be safe to use `to_atom`
- using `to_existing_atom` raises earlier than my valid option check, troublesome

princemaple:
@stevedomin please take another look.

stevedomin:
Thanks @princemaple, this is perfect!

PR #163, opened 2017-10-18
[WIP] Templates

anthonator:
Here's my take on templates for Swoosh. The implementation is very similar to Phoenix so it should be pretty familiar.

Here's example usage for the current implementation.

Note that the template engine used is determined by the extension of the view.

**lib/my_app/views/my_view.text.eex**
```
Hello, <%= @subject %>!
```

**lib/my_app/email.ex**
```elixir
defmodule MyApp.Email do
  use Swoosh.Template, root: "lib/my_app/views"

  import Swoosh.Email

  def foobar()
    new()
    |> to("anthonator@swoo.sh")
    |> from("partywagon@swoo.sh")
    |> subject("Swooooosh")
    |> text_body(render("my_view.text", %{ subject: "world" })
  end
end

MyApp.Email.foobar()
#> %Swoosh.Email{ text_body: "Hello, world!" }
```

I was thinking we could add `render_to_text` and `render_to_html` functions which would render and add the result to `text_body` and `html_body`. My initial thought was to add these to `Swoosh.Email`. What are your thoughts?

I need to be able to configure template engines. I was planning on just copying Phoenix for this.

```elixir
config :swoosh, :template_engines,
  eex: Swoosh.Template.EExEngine
```

Let me know your thoughts on progress so far.

- [x] Add behaviour for implementing a template engine
- [x] Add support for EEx
- [x] Add the ability to compile and render templates
- [x] Add the ability to configure template engines by extension
- [x] Add `render_to_text` and `render_to_html` functions
- [ ] More tests
- [ ] More documentation


stevedomin:
Hi @anthonator,

Sorry for the slow response, I'm travelling atm.

This looks like a great start. I'm going to have a more in-depth look tomorrow or after I'm back home on Wednesday.

Thank you so much for your work on this!

anthonator:
Here's an updated usage example.

**lib/my_app/email/templates/user/welcome.text.eex**
```
Welcome, <%= @name %>!

<%= body() %>
```

**lib/my_app/email/views/user.ex**
```elixir
defmodule MyApp.Email.Views.User do
  use Swoosh.Template, root: "lib/my_app/email/templates/user"

  def body() do
    "We hope you enjoy your stay!"
  end
end
```

**lib/my_app/email.ex**
```elixir
defmodule MyApp.Email do
  import MyApp.Email.Views.User
  import Swoosh.Email

  def send_welcome_email(user) do
    new()
    |> to(user.email)
    |> from("partywagon@swoo.sh")
    |> subject("Welcome!")
    |> assign(:name, user.name)
    |> render_to_text("welcome.text") # comes from MyApp.Email.Views.User
  end
end
```

```elixir
user = %MyApp.User{ email: "anthony@swoo.sh", name: "Anthony" }
MyApp.Email.send_welcome_email(user)
#=> %Swoosh.Email{ text_body: "Welcome, Anthony!\n\nWe hope you enjoy your stay!\n" }
```

anthonator:
@stevedomin have you had a chance to check this out yet?

stevedomin:
Hi @anthonator,

Yes I did. I need to pull the code and play with it a bit more but my main thought so far is that I'm wondering if there is any way we could make this a backward-compatible change for the people using the Phoenix integration?

What do you think?

anthonator:
Could you explain what you mean by "backward-compatible"?

stevedomin:
Hi @anthonator, I'm wayyyy too slow to reply these days.

Looks like you understood what I meant?

anthonator:
@stevedomin let me know what you think of what I have so far.

Here's an updated usage example:

**lib/my_app/email/templates/user/welcome.text.eex**
```elixir
Welcome, <%= @name %>!

<%= body() %>
```

**lib/my_app/email/views/user.ex**
```elixir
defmodule MyApp.Email.Views.User do
  use Swoosh.View, root: "lib/my_app/email/templates/user"

  def body() do
    "We hope you enjoy your stay!"
  end
end
```

```elixir
defmodule MyApp.Email do
  import Swoosh.Email

  def welcome_email(user) do
    new()
    |> to(user.email)
    |> from("partywagon@swoo.sh")
    |> subject("Welcome!")
    |> put_view(MyApp.Email.Views.User)
    |> assign(:name, user.name)
    |> render_to_text("welcome.text")
  end
end
```

```elixir
user = %MyApp.User{ email: "anthony@swoo.sh", name: "Anthony" }
MyApp.Email.welcome_email(user)
#=> %Swoosh.Email{ text_body: "Welcome, Anthony!\n\nWe hope you enjoy your stay!\n" }
```

stevedomin:
Hi @anthonator,

I've finally had time to deep diver into this. What you've done so far is great but I think there is a thing or two we should maybe move around.

Let's take a step back and talk about the API. Bear in mind these are just thoughts.

The only things `Swoosh.Email` should know about when it comes to rendering its html and text body are:

* `rendering_engine`: this is new
* `assigns`
* `private`
* `render_body/3`: this is new. The name can be changed but then it breaks compatibility with the existing `phoenix_swoosh` library.

`render_body/3` is a new function that takes the following params: `email`, `template`, `params`.

We can introduce a new behaviour `Swoosh.RenderingEngine` (better name?), that defines a callback `render_body/3`.

Swoosh would ship with a default EEx-based rendering engine, the one you already wrote. All the functions you recently added to `Swoosh.Email` would be pushed out of it to a new module that developers can `use` in their own email module (`MyApp.Email` in your case).

We can update the `phoenix_swoosh` package to implement the new behaviour.

It should also be quite easy to create a new package like `swoosh_smoothie` that implements the `Swoosh.RenderingEngine` behaviour.

What do you think? Would love to hear your comments, I'm sure there is plenty to improve.

anthonator:
It might make some sense to provide some reasoning behind the implementation thus far.

Templates are compiled at compile-time. `Swoosh.Template` compiles the EEx templates into the Elixir AST and creates `render/2` functions that can be called at run-time. The fact that only compiled Elixir is ever used gives a nice performance bump. This is also practically identical to how Phoenix templates work. It looks like Smoothie may operate similarly.

I also like this implementation because it allows the developer to separate the views from the email if they want. I like separation of concerns.

If you're thinking `Swoosh.RenderingEngine` would work similarly then we'd probably arrive at an identical solution to `Swoosh.Template` and `Swoosh.Template.Engine` but with different names. Personally, I like `Swoosh.Template` and `Swoosh.Template.Engine`, but naming is hard and I made it so... 😜

If you're thinking `Swoosh.RenderingEngine` would compile at runtime then you just need to be aware of the tradeoffs.

`Swoosh.View` exists to provide the ability to render nested templates. This gives us the concept of layouts. `Swoosh.View` uses `Swoosh.Template` under the hood with some additional logic. This is a slimmed down version of what Phoenix provides.

The functions I added to `Swoosh.Email` was an attempt to make it look like `phoenix_swoosh` to give you an idea of what that might look like. I prefer `render_to_text` and `render_to_html` because of its explicitness and it's non-assumption-ness but we don't have to get too deep into that right now.

Whether those functions live in `Swoosh.Email` or somewhere else doesn't matter to me. Adding them to `Swoosh.Email` may not make sense because it would conflict with functions on `phoenix_swoosh`.

Hopefully, this provides additional context.

stevedomin:
Sorry, I must not have explained my viewpoint in the right way.

I'm not proposing to change any of the stuff you did around template compilation, etc. This is great.

The only thing I'm advocating for is that we add another layer of indirection, because using a template system based on EEx is just one way of doing things.

You've done all the work on the "EEx templating" black box. What I'm thinking is that we also need a "Rendering" black box that would be completely agnostic to the way the email html and text body are constructed.

Does it make more sense? It not I'll put some code in the gist to better describe what I'm thinking.

anthonator:
I guess I am confused.

I'm not sure what the difference between `Swoosh.RenderingEngine` and `Swoosh.Template.Engine` would be.

The way it's been implemented isn't specific to EEx. I've just provided an EEx implementation (`Swoosh.Template.EExEngine`). Anyone can implement the `Swoosh.Template.Engine` behaviour and tell Swoosh to use it (e.g. `Swoosh.Template.SmoothieEngine`).

A gist would be helpful.

stevedomin:
First @anthonator, I'm really sorry about not getting my point in the right way. Really don't want to turn your contribution into a bad experience!

Back to the problem at hand.

Rendering might not be the right word in that case. But in my mind there is a difference between that "rendering engine" as I call it and the templating engine you have implemented.

`Swoosh.Template.Engine` is responsible for compiling template files into functions. Right now you've only implemented an EEx engine but we could use any templating language of course.

Now if we take Smoothie as an example, things are bit different. The way templates are compiled with Smoothie is slightly different and completely hidden to the developer. You just `use Smoothie` in a module and that module will get functions that returns an interpolated template string.

So there isn't really a way to implement the `Template.Engine` behaviour for Smoothie. This is just one example I imagine they might be others.

My point is, while the `Template.Engine` is perfect for solving the problem of "people might want to use different template engines" it does not solve the problem of "people might want to use a completely different mechanism for populating `html_body` and `text_body`".

Here's the kind of code I have in mind (pseudo-code):

**swoosh/lib/swoosh/rendering/engine.ex**
```elixir
defmodule Swoosh.Rendering.Engine do
  @callback render_body(email :: Swoosh.Email, rendering_key :: atom, assigns :: map) :: Swoosh.Email
end
```

**swoosh/lib/swoosh/rendering/engines/template_based.ex**
```elixir
defmodule Swoosh.Rendering.Engines.TemplateBased do
  def render_body(email, rendering_key, assigns) do
    email
     |> text_body(do_render(email, "#{rendering_key}.text", assigns))
     |> html_body(do_render(email, "#{rendering_key}.html", assigns))
  end

  def do_render(email, template, assigns) do
    layout  = Map.get(email.private, :swoosh_layout, false)
    module  = Map.get(email.private, :swoosh_view)
    
    assigns =
      assigns
      |> Enum.into(%{})
      |> Map.merge(email.assigns)
      |> Map.merge(%{ layout: layout })

    Swoosh.View.render(module, template, assigns)
  end
end
```

**swoosh/lib/swoosh/rendering/engines/smoothie.ex**
```elixir
defmodule Swoosh.Rendering.Engines.Smoothie do
  use Smoothie,
    template_dir: "templates",
    layout_file: Path.join("layouts", "main.html.eex")

  def render_body(email, rendering_key, assigns) when is_atom(rendering_key) do
   # smoothie will generate two functions inside that module, based on the templates present in the `templates` directory (example: `welcome_text` and `welcome_html`)
    text_body = apply(__MODULE__,  "#{rendering_key}_text", [assigns])
    html_body = apply(__MODULE__,  "#{rendering_key}_html", [assigns])

    email
     |> text_body(text_body)
     |> html_body(html_body)
  end

  ...
end
```

**swoosh/lib/swoosh/email.ex**
```elixir
defmodule Swoosh.Email do
  ...

  def render_body(email, rendering_key, assigns) do
    rendering_engine = email.rendering_engine || Application.get(:swoosh, :default_rendering_engine)
    
    rendering_engine.render_body(email, rendering_key, assigns)
  end
end
```

**lib/my_app/email.ex**
```elixir
defmodule MyApp.Email do
  import MyApp.Email.Views.User
  import Swoosh.Email

  def welcome_email(user) do
    new()
    |> to(user.email)
    |> from("partywagon@swoo.sh")
    |> subject("Welcome!")
    |> assign(:name, user.name)
    |> render_body(:welcome, %{})
  end
end
```

princemaple:
@anthonator great PR :) finally got around to take a look

@stevedomin looks like you simply want the `do_render` part extracted out to another module, which represents the extra layer of abstraction, plus making it configurable?

stevedomin:
@princemaple yes, I guess that's the gist of it :)

stevedomin:
Hi @anthonator, just checking that you are still keen on tackling this one? Let us know if we can do anything to help!

I know I spent a lot of time discussing and it might have been a bit frustrating for you but I'm really keen to get this merged.

anthonator:
Sorry, I've been swamped and I don't have the resources to spend more time on this.

My original goal was to create an adapter based template engine for Swoosh. Adding the ability to support multiple template engines is beyond the scope of what I was trying to accomplish. I think that's why I've been confused.

@stevedomin, I think it's obvious that you have a specific vision of what you'd like to see here. I think that's great and if you'd like to take this across the finish line that's perfectly OK with me. If you'd rather I make this a separate project that can plug into some sort of generic rendering layer I'd be willing to do that as well.

I haven't been frustrated so no hard feelings. I just have a limited amount of time.

stevedomin:
@anthonator completely understandable, I'm sorry for the confusion.

I'll take this over the finish line and if you can spare some time I'll tag you for feedback.

Thanks again

princemaple:
Changed merge target into `template` branch I just created. Merging now so I can easily work on it when I get some time.

PR #166, opened 2017-10-23
Enforce string values

princemaple:


princemaple:
@anthonator You might be interested in taking a look a this.

princemaple:
The thing with ignoring nil is, if you make a mistake in prod and you don't see an immediate explosion, you are going to find out about it in a much more unpleasant way...

stevedomin:
@anthonator thought the same but actually I kinda like the explicitness brought by not ignoring nils. You always have a username, password and relay so even it's empty at least you've consciously made the decision when creating your config with env vars or whatever.

PR #167, opened 2017-10-25
Amazon SES Adapter

jdollar:
Attempt at adding a new Amazon SES adapter to send out basic emails. Uses their SendRawEmail interface on their query api since that is the interface that allows for attachments.

Addresses #6

### DIFF
--- a/lib/swoosh/adapters/amazon_ses.ex
+++ b/lib/swoosh/adapters/amazon_ses.ex
@@ -1,0 +1,257 @@
+defmodule Swoosh.Adapters.AmazonSes do
+  @moduledoc ~S"""
+  An adapter that sends email using the Amazon Simple Email Service Query API.
+
+  For reference: [Amazone SES Query Api Docs](http://docs.aws.amazon.com/ses/latest/APIReference/Welcome.html)
+
+  ## Example
+
+      # config/config.exs
+      config :sample, Sample.Mailer,
+        adapter: Swoosh.Adapters.AmazonSes,
+        region: 'region-endpoint',
+        access_key: 'aws-access-key',
+        secret: 'aws-secret-key'
+
+      # lib/sample/mailer.ex
+      defmodule Sample.Mailer do
+        use Swoosh.Mailer, otp_app: :sample
+      end
+  """
+
+  use Swoosh.Adapter, required_config: [:region, :access_key, :secret]
+  alias Swoosh.Email
+  alias Swoosh.Adapters.XML.Helpers, as: XmlHelper
+  alias Swoosh.Adapters.SMTP.Helpers, as: SMTPHelper
+
+  @encoding     "AWS4-HMAC-SHA256"
+  @host_prefix  "email."
+  @host_suffix  ".amazonaws.com"
+  @service_name "ses"
+  @action       "SendRawEmail"
+  @base_headers %{"Content-Type" => "application/x-www-form-urlencoded"}
+  @version      "2010-12-01"
+
+  def deliver(%Email{} = email, config \\ []) do
+    body = email |> prepare_body(config) |> encode_body
+    url = base_url(config)
+    headers = prepare_headers(@base_headers, body, config)
+
+    case :hackney.post(url, headers, body, [:with_body]) do
+      {:ok, 200, _headers, body} -> {:ok, interpret_response(body)}
+      {:ok, code, _headers, body} when code > 399 ->
+        {:error, interpret_error_response(body)}
+      {:error, reason} -> {:error, reason}
+    end
+  end
+
+  defp interpret_response(body) do
+    id =
+      body
+      |> XmlHelper.parse
+      |> XmlHelper.first("//MessageId")
+      |> XmlHelper.text
+
+    %{id: id}
+  end
+
+  defp interpret_error_response(body) do
+    IO.inspect body
+    node =
+      body
+      |> XmlHelper.parse
+
+    code =
+      node
+      |> XmlHelper.first("//Error/Code")
+      |> XmlHelper.text
+
+    message =
+      node
+      |> XmlHelper.first("//Error/Message")
+      |> XmlHelper.text
+
+    %{code: code, message: message}
+  end
+
+  defp base_url(config) do
+    case config[:host] do
+      nil -> "https://" <> @host_prefix <> config[:region] <> @host_suffix
+      _ -> config[:host]
+    end
+  end
+
+  defp version(config) do
+    case config[:version] do
+      nil -> @version
+      _ -> config[:version]
+    end
+  end
+
+  defp prepare_body(email, config) do
+    raw_body =
+      SMTPHelper.body(email, config)
+      |> Base.encode64
+      |> URI.encode_www_form
+
+    %{}
+    |> Map.put("Action", @action)
+    |> Map.put("Version", version(config))
+    |> Map.put("RawMessage.Data", raw_body)
+    # |> prepare_source(from)
+    # |> prepare_recipients(to)
+    # |> prepare_recipients(cc, "CcAddresses")
+    # |> prepare_recipients(bcc, "BccAddresses")
+    # |> prepare_reply_to(email)
+    # |> prepare_subject(email)
+    # |> prepare_content(email)
+  end
+
+  defp prepare_content(body, email) do
+    body
+    |> prepare_html(email)
+    |> prepare_text(email)
+  end
+
+  defp prepare_content(body, _), do: body
+
+  defp prepare_source(body, {"", from_email}), do: Map.put(body, "Source", from_email)
+  defp prepare_source(body, {from_name, from_email}) do
+    Map.put(body, "Source", "\"#{from_name}\" <#{from_email}>")
+  end
+
+  defp prepare_recipients(body, emails, type \\ "ToAddresses", count \\ 1)
+  defp prepare_recipients(body, [{name, address} | tail], type, count) when name == "" or name == nil do
+    Map.put(body, "Destination.#{type}.member.#{count}", address)
+    |> prepare_recipients(tail, type, count + 1)
+  end
+
+  defp prepare_recipients(body, [{name, address} | tail], type, count) do
+    Map.put(body, "Destination.#{type}.member.#{count}", "\"#{name}\" <#{address}>")
+    |> prepare_recipients(tail, type, count + 1)
+  end
+
+  defp prepare_reply_to(body, %{reply_to: nil}), do: body
+  defp prepare_reply_to(body, %{reply_to: reply_to}), do: prepare_reply_to(body, reply_to)
+  defp prepare_reply_to(body, [{_name, address} | tail], count \\ 0) do
+    body
+    |> Map.put("ReplyToAddresses.member.#{count}", address)
+    |> prepare_reply_to(tail, count)
+  end
+  defp prepare_reply_to(body, [], _), do: body
+
+  defp prepare_recipients(body, [], _, _), do: body
+
+  defp encode_body(body) do
+    Map.keys(body)
+    |> Enum.sort
+    |> Enum.map(fn(key) -> "#{key}=#{body[key]}" end)
+    |> Enum.join("&")
+    |> URI.encode
+  end
+
+  defp prepare_headers(headers, body, config) do
+    signed_header_list = "content-type;host;x-amz-date"
+    current_date_time = DateTime.utc_now()
+
+    headers =
+      headers
+      |> prepare_header_host(config)
+      |> prepare_header_date(config, current_date_time)
+
+    headers_string = setup_headers_string(headers)
+    signature =
+      determine_request_hash(body, headers_string, signed_header_list)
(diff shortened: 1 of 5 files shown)


princemaple:
Hi @jdollar,

Thanks for the great looking PR!
I'll give it a thorough review soon.

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,229 @@
...
+  defp encrypt_value(secret, unencrypted_data), do: :crypto.hmac(:sha256, secret, unencrypted_data)
+
+  defp extract_date_string(dt), do: "#{dt.year}#{dt.month}#{dt.day}"
+  defp normalize_iso8601(dt) do
+    date = extract_date_string(dt)
+    {_, time} = Time.new(dt.hour, dt.minute, dt.second)
+    time_string =
+      time
+      |> Time.to_string 
+      |> String.replace(":", "")
```
princemaple: You did `"#{dt.year}#{dt.month}#{dt.day}"` for date, why not just `"#{dt.hour}#{dt.minute}#{dt.second}"` here as well? `dt` is generated by `DateTime.utc_now` anyway, it couldn't be a bad value.

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,229 @@
...
+      "#{date}/#{config[:region]}/#{@service_name}/aws4_request",
+      request_hash
+    ]
+    |> Enum.join("\n")
+  end
+
+  defp encrypt_value(secret, unencrypted_data), do: :crypto.hmac(:sha256, secret, unencrypted_data)
+
+  defp extract_date_string(dt), do: "#{dt.year}#{dt.month}#{dt.day}"
+  defp normalize_iso8601(dt) do
```
princemaple: Let's make this two clear that they are Amazon specific. I was confused in why you don't simply `Date.to_string` and `NaiveDateTime.to_string`.

Maybe `amz_date` and `amz_datetime`? Since it's called `X-Amz-Date` in the header.

lib/swoosh/adapters/xml/helpers.ex
```diff
@@ -0,0 +1,37 @@
...
+
+    node
+  end
+
+  @doc false
+  def first(node, path) do
+    path
+    |> to_char_list
+    |> :xmerl_xpath.string(node)
+    |> extract_first
```
princemaple: `hd` or `List.first`

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,229 @@
...
+
+  defp interpret_error_response(body) do
+    node =
+      body
+      |> XmlHelper.parse
+
+    code =
+      node
+      |> XmlHelper.first("//Error/Code")
+      |> XmlHelper.text
```
princemaple: If these two functions are only ever going to be used together, maybe merge them?

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,229 @@
...
+    authorization = prepare_authorization(config, signed_header_list, current_date_time, signature)
+    Map.put(headers, "Authorization", authorization)
+  end
+
+  defp setup_headers_string(headers) do
+    headers
+    |> Map.keys
+    |> Enum.sort
+    |> Enum.map(fn(key) -> "#{String.downcase(key)}:#{headers[key]}" end)
+    |> Enum.join("\n")
```
princemaple: `headers |> Enum.sort |> Enum.map_join("\n", fn {k, v} -> "#{String.downcase(k)}:#{v}" end)`

lib/swoosh/adapters/amazon_ses.ex
jdollar: That sounds good to me. I'll update them to make it more obvious.

Kind of did want to draw attention to the NaiveDateTime.to_iso8601(native_date_time, :basic) and Date.to_iso8601(date, :basic) functions that were added in the 1.5 version of elixir that would essentially do the same thing. I believe with the current 1.4 version it doesn't have the option to specify what iso8601 format you want to use which is why I had to fall back to these two functions.

lib/swoosh/adapters/amazon_ses.ex
jdollar: I believe the reason I went about creating the time object was that the to_string value kept the 0 padding at the beginning of single digit fields. So it came out to 02:10:20 with the Time.to_string and replace vs 2:10:20 with the `#{dt.hour}#{dt.minute}#{dt.second}`. If I remember right, Amazon was yelling at me when I tried to send up a datetime would that zero padding at the beginning.

jdollar: Tried a little something different here. Used the `String.pad_leading` function instead of making that new time object and going that route.

jdollar:
Thanks for running through it @princemaple. I appreciate all the elixir knowledge bombs you were dropping on me. Made some updates based on the comments!

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+
+  defp encrypt_value(secret, unencrypted_data), do: :crypto.hmac(:sha256, secret, unencrypted_data)
+
+  defp amz_date(dt), do: "#{dt.year}#{dt.month}#{dt.day}"
+  defp amz_datetime(dt) do
+    time_string = Enum.join([
+      String.pad_leading(Integer.to_string(dt.hour), 2, "0"),
+      String.pad_leading(Integer.to_string(dt.minute), 2, "0"),
+      String.pad_leading(Integer.to_string(dt.second), 2, "0")
+    ])
```
princemaple: FWIW `Enum.map_join([dt.hour, dt.minute, dt.second], &String.pad_leading(to_string(&1), 2, "0"))` :)

test/swoosh/adapters/amazonses_test.exs
```diff
@@ -0,0 +1,127 @@
...
+      region: "us-east-1",
+      access_key: "test_access",
+      secret: "test_secret"
+    ]
+
+    valid_email =
+      new()
+      |> from("guybrush.threepwood@pirates.grog")
+      |> to("elaine.marley@triisland.gov")
+      |> subject("Mighty Pirate Newsletter")
```
princemaple: Woah fancy, what are they? 😃

test/swoosh/adapters/xml/helpers_test.exs
```diff
@@ -0,0 +1,57 @@
+defmodule Swoosh.Adapters.XML.HelpersTest do
+  use ExUnit.Case, async: true
+  alias Swoosh.Adapters.XML.Helpers, as: XMLHelper
+
+  setup do
```
princemaple: this can probably be a `setup_all`

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
+defmodule Swoosh.Adapters.AmazonSes do
```
stevedomin: I'd go with AmazonSES as the module name

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+  @version      "2010-12-01"
+
+  def deliver(%Email{} = email, config \\ []) do
+    query = email |> prepare_body(config) |> encode_body
+    url = base_url(config)
+    headers = prepare_headers(@base_headers, query, config)
+
+    case :hackney.post(url, headers, query, [:with_body]) do
+      {:ok, 200, _headers, body} ->
+        {:ok, interpret_response(body)}
```
stevedomin: Completely subjective but I'd prefer `parse_response` here, seems more "standard".

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+      {:ok, code, _headers, body} when code > 399 ->
+        {:error, interpret_error_response(body)}
```
stevedomin: Same here, I'd go with `parse_error_response`

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+  end
+
+  defp interpret_response(body) do
+    id =
+      body
+      |> XMLHelper.parse
+      |> XMLHelper.first("//MessageId")
+      |> XMLHelper.text
+
+    %{id: id}
```
stevedomin: Is there anything else useful in that response body? If there is, feel free to add it to the result map.

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+  end
+
+  defp interpret_error_response(body) do
+    node = body |> XMLHelper.parse
```
stevedomin: Pipeline operator is not needed here.

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+
+    code = node |> XMLHelper.first_text("//Error/Code")
```
stevedomin: Pipeline operator is not needed here.

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+    message = node |> XMLHelper.first_text("//Message")
```
stevedomin: Pipeline operator is not needed here.

lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -0,0 +1,203 @@
...
+  defp setup_headers_string(headers) do
+    headers
+    |> Enum.sort
+    |> Enum.map_join("\n", fn {k, v} -> "#{String.downcase(k)}:#{v}" end)
+  end
+
+  defp generate_signed_header_list(headers) do
+    headers
+    |> Map.keys
+    |> Enum.map(fn(key) -> String.downcase(key) end)
```
stevedomin: You can use `Enum.map_join/3` here as well.

lib/swoosh/adapters/amazon_ses.ex
princemaple: I was going to wait for 1.6 formatter to rule them all 😈

lib/swoosh/adapters/amazon_ses.ex
princemaple: also `&String.downcase/1`

lib/swoosh/adapters/amazon_ses.ex
stevedomin: 👍

jdollar:
Thanks for combing though everything. Made some updates after going through the comments out there. As requested too, I made some updates to the documentation. Included some basic details on the README in regards to the gen_smtp requirements, but I put a bit more information on it in the actual moduledoc annotation block. Does that look cool? Did we want to stick that somewhere else or provide some different information on it?

test/swoosh/adapters/amazonses_test.exs
jdollar: The mightiest pirate who sailed the seas and his lass! A couple names from a game series a while ago. https://en.wikipedia.org/wiki/The_Secret_of_Monkey_Island

Definite recommend if you like those point and click kind of games. Think they remade a couple of them recently too. Like to dip my toes into my nostalgia when I can I guess.

test/swoosh/adapters/xml/helpers_test.exs
jdollar: Was able to shift some of it out of the setup block. Had to keep the setup block around to run `Bypass.open` before every test and setup the host, but I was able to extract a majority of the config and the email generation out to the setup_all block.

princemaple: We were probably talking about different `setup` blocks here :) I'm so glad that you went and cleaned up the other block, though!

stevedomin: @jdollar this `setup` block can definitely be changed for a `setup_all`

stevedomin:
@jdollar one last thing to change and I'll merge this. Thank you so much.

test/swoosh/adapters/xml/helpers_test.exs
jdollar: @stevedomin @princemaple Yeah! You were right. I was thinking this was referring the setup block in the amazon SES test file originally. My bad. My selective reading kicked in or something. Update should be out there now.

princemaple:
@stevedomin is this one good to 🚢 now?

jdollar:
Thanks for the help getting this out there @princemaple @stevedomin. 👍 on the PR experience.

princemaple:
Glad that you are OK with my nitpicks! :)

stevedomin:
Thank you so much for your work @jdollar!

PR #169, opened 2017-10-30
Remove redundant decoding

### LINKED ISSUE #168: Mandrill adapter, ** (ArgumentError) argument error (and fix)
We run out of mails in the test plan on Mandrill and it started rejecting messages with `"status": "invalid"`. This is normal but the adapter throws this error

```
     ** (ArgumentError) argument error
     stacktrace:
       :erlang.iolist_to_binary(%{"_id" => "...", "email" => "mymail@example.com", "reject_reason" => nil, "status" => "invalid"})
       (poison) lib/poison/parser.ex:35: Poison.Parser.parse/2
       (poison) lib/poison/parser.ex:51: Poison.Parser.parse!/2
       (poison) lib/poison.ex:83: Poison.decode!/2
       (swoosh) lib/swoosh/adapters/mandrill.ex:46: Swoosh.Adapters.Mandrill.interpret_response/1
       ...
```

Line 46 is the last one of

```
  defp interpret_response(body) when is_binary(body), do: body |> Poison.decode! |> hd |> interpret_response
  defp interpret_response(%{"status" => "sent"} = body), do: {:ok, %{id: body["_id"]}}
  defp interpret_response(%{"status" => "queued"} = body), do: {:ok, %{id: body["_id"]}}
  defp interpret_response(%{"status" => "rejected"} = body), do: {:error, body}
  defp interpret_response(body), do: {:error, Poison.decode!(body)}
```

The error is caused because `body`, which is a map, is eventually passed to `:erlang.iolist_to_binary` which expects a binary and not a map.

The fix is not to `decode!` `body`, because it's already decoded:

```
  defp interpret_response(body), do: {:error, body}
```

princemaple:
fixes #168 

also syncs function name with the suggestion in #167

### DIFF
--- a/lib/swoosh/adapters/mandrill.ex
+++ b/lib/swoosh/adapters/mandrill.ex
@@ -31,7 +31,7 @@ defmodule Swoosh.Adapters.Mandrill do
 
     case :hackney.post(url, @headers, body, [:with_body]) do
       {:ok, 200, _headers, body} ->
-        interpret_response(body)
+        parse_response(body)
       {:ok, code, _headers, body} when code > 399 ->
         {:error, {code, Poison.decode!(body)}}
       {:error, reason} ->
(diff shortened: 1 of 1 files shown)


stevedomin:
thanks @princemaple!!!

PR #175, opened 2017-11-13
Add support for send-template endpoint

### LINKED ISSUE #174: Add support for send-template endpoint
when sending an email using a template we need to hit a different endpoint `/messages/send-template.json` also we need to specify a `template_content` for the sending template and that is obligatory but can be blank.

sescobb27:
when sending an email using a template we need to hit a different endpoint `/messages/send-template.json` also we need to specify a `template_content` for the sending template and that is obligatory but can be blank.

closes #174

### DIFF
--- a/lib/swoosh/adapters/mandrill.ex
+++ b/lib/swoosh/adapters/mandrill.ex
@@ -23,11 +23,12 @@ defmodule Swoosh.Adapters.Mandrill do
 
   @base_url     "https://mandrillapp.com/api/1.0"
   @api_endpoint "/messages/send.json"
+  @template_api_endpoint "/messages/send-template.json"
   @headers      [{"Content-Type", "application/json"}]
 
   def deliver(%Email{} = email, config \\ []) do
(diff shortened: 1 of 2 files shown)


lib/swoosh/adapters/mandrill.ex
```diff
@@ -47,10 +48,15 @@ defmodule Swoosh.Adapters.Mandrill do
 
   defp base_url(config), do: config[:base_url] || @base_url
 
+  defp api_endpoint(%{provider_options: %{template_name: _template_name}}), do: @template_api_endpoint
+  defp api_endpoint(%{provider_options: %{template_content: _template_content}}), do: @template_api_endpoint
```
stevedomin: Reading the API docs it looks like `template_name` is mandatory if you want to use templates so we could use this to figure out which endpoint we need to use.

sescobb27:
@stevedomin done :)

stevedomin:
Thanks for the PR @sescobb27!

princemaple:
@sescobb27 released in 0.12 FYI

PR #176, opened 2017-11-15
add optional headers to attachment

davec82:
When you add an attachment, sometimes, you need to add some custom headers (i.e. for a ical). This PR add a (optional) list of custom headers when a new attachment struct is created.


stevedomin:
Thanks for the PR @davec82, we definitely wanted that!

PR #178, opened 2017-11-16
add in memory attachment for smtp adapter

davec82:
Added "data" key in Swoosh.Attachment so you can add "in memory" data as attachment. Added this feature for SMTP adapter.

### DIFF
--- a/lib/swoosh/adapters/smtp/helpers.ex
+++ b/lib/swoosh/adapters/smtp/helpers.ex
@@ -102,21 +102,24 @@ if Code.ensure_loaded?(:mimemail) do
        content}
     end
 
-    defp prepare_attachment(%{filename: filename, path: path, content_type: content_type, type: attachment_type, headers: custom_headers}) do
+    defp prepare_attachment(%{filename: filename, path: path, content_type: content_type, type: attachment_type, headers: custom_headers, data: data}) do
       [type, format] = String.split(content_type, "/")
-      file = File.read!(path)
+      content = case data do
+        nil -> File.read!(path)
+        _   -> data
+      end
 
       case attachment_type do
         :attachment -> {type, format,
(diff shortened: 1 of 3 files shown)


lib/swoosh/adapters/smtp/helpers.ex
```diff
@@ -102,21 +102,24 @@ if Code.ensure_loaded?(:mimemail) do
        content}
     end
 
-    defp prepare_attachment(%{filename: filename, path: path, content_type: content_type, type: attachment_type, headers: custom_headers}) do
+    defp prepare_attachment(%{filename: filename, path: path, content_type: content_type, type: attachment_type, headers: custom_headers, data: data}) do
       [type, format] = String.split(content_type, "/")
-      file = File.read!(path)
+      content = case data do
```
stevedomin: Shall we have this logic in the `Attachment` module? Otherwise we'll end up duplicating it in all adapters.

Maybe we can populate `data` when the attachment is created?

A lot of our adapters are doing `path |> File.read! |> Base.encode64` to include the attachments in the request body.

davec82: Do you mean adding this logic to new function in `Attachment`? Anyway you can create attachment using `%Swoosh.Attachment{content_type: "text/plain", data: "text text"}` but in this case how can you add the logic?

stevedomin: Good point. Maybe just a function that takes an `Attachment` and returns the content for it then?

davec82: Added a new function to `Attachment` module that retrieves content from struct

stevedomin:
Thanks again for the PR @davec82!!

princemaple:
Released in 0.12 FYI

PR #180, opened 2017-11-21
Fix dialyzer warnings when using Swoosh.Mailer

xadhoom:
As soon as a module uses Swoosh.Mailer, dialyzer warns about:

> lib/my_mailer.ex:3: The pattern {'error', _@5, _@6} can never match the type {'error',_} | {'ok',_}

This happens because the case into mailer.ex looks for a `triple` which never happens in any of the adapters. As confirmation the deliver callback typespec does not specify such triple.

This PR fixes that and also fixes the return value from the sendmail.ex deliver fun, because Port.close returns :true while the specs expects at least :ok

### DIFF
--- a/lib/swoosh/adapters/sendmail.ex
+++ b/lib/swoosh/adapters/sendmail.ex
@@ -28,6 +28,7 @@ if Code.ensure_loaded?(:mimemail) do
       port = Port.open({:spawn, cmd(email, config)}, [:binary])
       Port.command(port, body)
       Port.close(port)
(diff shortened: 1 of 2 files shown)


princemaple:
Hi @xadhoom Thanks for the PR!

princemaple:
@stevedomin This PR does not seem to conflict with your error refactor PR, so I'm merging it in.

princemaple:
Integration test is still failing

Issue #181, opened 2017-11-22
Provide Attachment.new for in-memory data


princemaple:
Alternatively, separate everything into
- `Attachment.from_plug_conn`
- `Attachment.from_data`
- `Attachment.from_path`

@stevedomin thoughts?

stevedomin:
I like the idea of handling data in `Attachment.new` but I'm not sure how we can differentiate between that and a path, they're both binary no?

princemaple:
We could expand the current functionality with tagged data like `Attachment.new({:data, binary}, opts)`
That being said, I prefer my other options listed above. If needed, add `new` equivalents as well.

stevedomin:
Sounds good!

Should we mark `Attachment.new/1` as deprecated and update the docs to use the functions above?

stevedomin:
We can deprecate in 1.0

princemaple:
Yeah we could. Will be a long journey though :)

Doerge:
#288 works with Mandrill as well 🎉 
```elixir
|> attachment(Attachment.new(
      {:data, binary_data},
      filename: "foobar.pdf",
      content_type: "application/pdf")
    )
```

ckruse:
I started using swoosh and stumbled upon this issue. It looks like the `{:data, binary}` code already exists:

~~~ elixir
  def new({:data, data}, opts) do
    %{struct!(__MODULE__, opts) | data: data}
  end
~~~

And it works like a charm! 👍 Sadly the spec for `new/2` is wrong:

~~~elixir
  @spec new(binary | struct, Keyword.t() | map) :: %__MODULE__{}
~~~

It has to be something like this:

~~~elixir
  @spec new(binary | struct | {:data, binary}, Keyword.t() | map) :: %__MODULE__{}
~~~

princemaple:
Hi @ckruse , would you like to send a PR to correct that? 😃

ckruse:
@princemaple Of course! ✅done: #413

princemaple:
forgot to say thanks @ckruse

PR #182, opened 2017-11-22
Refactor to use get_content in all adapters

### LINKED ISSUE #179: Add support for in-memory attachment to all adapters
https://github.com/swoosh/swoosh/pull/178 added support for in-memory attachments.

We should use the newly introduced `Attachment.get_content/1` in all adapters that supports it.

### LINKED ISSUE #183: Incomplete typespec in Swoosh.Attachment.new
https://github.com/swoosh/swoosh/blob/cee805dc626b2f2e7692c18f5cc72371547ec35a/lib/swoosh/attachment.ex#L28

it should be `@spec new(binary | struct, Keyword.t) :: %__MODULE__{}` since it accepts also a Plug.Upload struct

princemaple:
Close #179 
Fixes #183 

This PR will allow sending attachments from in-memory data in all adapters.
However, we only provide one convenience function `Attachment.new` that works with `path`, and `Plug.Conn` if it's included in the code base. I filed a separate issue #181 for further improvements that will allow in-memory data and maybe other ways of adding attachments in the future.

### DIFF
--- a/lib/swoosh/attachment.ex
+++ b/lib/swoosh/attachment.ex
@@ -75,13 +75,7 @@ defmodule Swoosh.Attachment do
     raise Swoosh.AttachmentContentError, message: "No path or data is provided"
   end
   def get_content(%__MODULE__{data: data, path: path}, encoding \\ :raw) do
-    content =
-      case data do
-        nil -> File.read!(path)
-        _   -> data
-      end
-
-    encode(content, encoding)
+    encode(data || File.read!(path), encoding)
   end
 
   defp encode(content, :raw), do: content
--- a/lib/swoosh/attachment.ex
+++ b/lib/swoosh/attachment.ex
@@ -3,9 +3,16 @@ defmodule Swoosh.Attachment do
   Struct representing an attachment in an email.
   """
 
-  defstruct filename: nil, content_type: nil, path: nil, type: nil, headers: [], data: nil
+  defstruct [:filename, :content_type, :path, :type, :headers, :data]
 
-  @type t :: %__MODULE__{}
+  @type t :: %__MODULE__{
+    filename: String.t,
+    content_type: String.t,
+    path: String.t,
+    type: :inline | :attachment,
+    headers: [{String.t, String.t}],
+    data: binary
+  }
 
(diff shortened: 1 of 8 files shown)


princemaple:
@xadhoom I added more typespecs in this PR, would you like to review it?

princemaple:
SMTP integration tests are failing

lib/swoosh/attachment.ex
```diff
@@ -3,9 +3,16 @@ defmodule Swoosh.Attachment do
...
+  defstruct [:filename, :content_type, :path, :type, :headers, :data]
 
-  @type t :: %__MODULE__{}
+  @type t :: %__MODULE__{
+    filename: String.t,
+    content_type: String.t,
+    path: String.t,
+    type: :inline | :attachment,
+    headers: [{String.t, String.t}],
+    data: binary
```
xadhoom: according to `get_content/2` data should be `binary | nil`

princemaple: Then `path` should be allowed to be `nil` too, right? When it's in-memory data.

princemaple: Thanks for the review BTW!

xadhoom: good catch.

That brings me another little optimization, since data and path cannot be nil at the same time.
You can add a function clause like:

`def get_content(%__MODULE__{data: nil, path: nil})`

that spits out an exception or {:error, term}, or anything like, just to be sure user is alerted by your layer (otherwise a File.read! exception will raise)

princemaple: Right. I'll add a new error for that. Thanks @xadhoom .

stevedomin:
This looks great @princemaple!

princemaple:
@xadhoom @stevedomin updated with an addition of `Swoosh.AttachmentContentError`. It will need to be merged into `exceptions.ex` when error unification PR is done. Please take another look.

princemaple:
OK the failed build actually did catch a bug.

xadhoom:
Looks fine to me, nice!

PR #184, opened 2017-11-27
Fix text/html alternative multipart when having also an attachment

xadhoom:
Right now, if both text and html are given along an attachment, the smtp adapter builds a single level multipart/mixed , resulting into text and html shown at the same time.

Instead, html and text must be a multipart/alternative *inside* the multipart/mixed which holds the attachment.

This PR fixes that.

### DIFF
--- a/lib/swoosh/adapters/smtp/helpers.ex
+++ b/lib/swoosh/adapters/smtp/helpers.ex
@@ -67,50 +67,42 @@ if Code.ensure_loaded?(:mimemail) do
     end
 
     defp prepare_parts(headers, %{
-      attachments: attachments,
-      html_body: nil,
-      text_body: text_body
-    }) when length(attachments) > 0 do
-      parts = Enum.map(attachments, &prepare_attachment(&1))
-      parts = [prepare_part(:plain, text_body) | parts]
-
-      {"multipart", "mixed", headers, parts}
-    end
-    defp prepare_parts(headers, %{
-      attachments: attachments,
+      attachments: [],
       html_body: html_body,
-      text_body: nil
-    }) when length(attachments) > 0 do
-      parts = Enum.map(attachments, &prepare_attachment(&1))
-      parts = [prepare_part(:html, html_body) | parts]
-
-      {"multipart", "mixed", headers, parts}
+      text_body: text_body
+    }) do
+      case {text_body, html_body} do
+        {text_body, nil} ->
+          headers = [{"Content-Type", "text/plain; charset=\"utf-8\""} | headers]
+          {"text", "plain", headers, text_body}
+        {nil, html_body} ->
+          headers = [{"Content-Type", "text/html; charset=\"utf-8\""} | headers]
+          {"text", "html", headers, html_body}
+        {text_body, html_body} ->
+          parts = [prepare_part(:plain, text_body), prepare_part(:html, html_body)]
+          {"multipart", "alternative", headers, parts}
+      end
     end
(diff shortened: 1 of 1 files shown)


princemaple:
Hi @xadhoom , thanks again for another PR!

The multipart logic has grown lot recently since the addition of inline attachments, in-memory attachments and custom attachment headers. I pushed a refactor commit directly onto your branch. Could you take a look and tell me if you are happy with it?

lib/swoosh/adapters/smtp/helpers.ex
```diff
@@ -102,21 +115,44 @@ if Code.ensure_loaded?(:mimemail) do
...
-        :attachment -> {type, format,
-                         [{"Content-Transfer-Encoding", "base64"}] ++ custom_headers,
-                         [{"disposition", "attachment"}, {"disposition-params", [{"filename", filename}]}],
-                         content}
-        :inline     -> {type, format,
-                         [{"Content-Transfer-Encoding", "base64"}, {"Content-Id", "<#{filename}>"}] ++ custom_headers,
-                         [{"content-type-params", []},
-                          {"disposition", "inline"},
-                          {"disposition-params", []}],
-                         content}
```
princemaple: only indentation changes here.

lib/swoosh/adapters/smtp/helpers.ex
```diff
@@ -102,21 +115,44 @@ if Code.ensure_loaded?(:mimemail) do
        content}
     end
 
-    defp prepare_attachment(%{filename: filename, content_type: content_type, type: attachment_type, headers: custom_headers} = attachment) do
+    defp prepare_attachment(%{
+      filename: filename,
+      content_type: content_type,
+      type: attachment_type,
+      headers: custom_headers
+    } = attachment) do
```
princemaple: only indentation changes here.

lib/swoosh/adapters/smtp/helpers.ex
```diff
@@ -67,50 +67,42 @@ if Code.ensure_loaded?(:mimemail) do
...
+            html_part
+          {text_part, html_part} ->
+            {"multipart", "alternative", [], [], [text_part, html_part]}
+        end
+      attachment_parts = Enum.map(attachments, &prepare_attachment(&1))
+
+      {"multipart", "mixed", headers, [content_part | attachment_parts]}
     end
 
+    defp prepare_part(_subtype, nil), do: nil
```
xadhoom: maybe is worth checking also for the empty string? just in case of poor usage by the user...

stevedomin:
Thanks for the PR @xadhoom, can't believe we had this bug for so long!

lib/swoosh/adapters/smtp/helpers.ex
stevedomin: Might be completely wrong here but I think we're only checking `nil` so that user can still have empty parts (allowed by the spec).

princemaple: Hmm, I'm a little skeptical about changing behavior in this PR. Though not probable, someone could be using it as a feature 😉

xadhoom: sure, if the standard allows that, I'm fine with it :)

princemaple:
@xadhoom this is out in 0.12 FYI

xadhoom:
@princemaple thanks! I was already using master btw :smile:

PR #189, opened 2017-12-18
Return the message id from the sendgrid adapter mirroring other adapters

felipesere:
See
https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html#-Response-Example
for the description of the different headers.


princemaple:
Might be better to extract the header part in test into a function. Otherwise LGTM.

princemaple:
FYI I'm visiting family in China, and Steve was travelling too last time we talked. Forgive us if there isn't much communication in the next few days/weeks! Oh, and thanks for the contribution!

princemaple:
I'll take care of this soon.

princemaple:
Thanks @felipesere !

PR #194, opened 2018-01-10
Mailgun provider_option handling recipient variables

messutied:
I also tested this with sending actual emails with Mailgun and it works.


princemaple:
Looks good, thanks!

princemaple:
Looks good, thanks!

PR #201, opened 2018-01-31
chore: added explanation for Swoosh.Attachment struct

DKhalil:
This should make it clearer what is possible with the attachments in swoosh


stevedomin:
Thank @DKhalil!

PR #203, opened 2018-02-13
Add no_mx_lookups to smtp config

sergioaugrod:
Hello,

I added `no_mx_lookups` option to `smtp` config.
This option can help when you don't want to connect to the returned mx's of relay.

The `no_mx_lookups` is not mentioned in the `gen_smtp` README, but see https://github.com/Vagabond/gen_smtp/blob/master/src/gen_smtp_client.erl#L136 for more details.

It's now possible to pass this option, but making it clearer in the documentation can help other people.

What do you think? Would it be valid?

Thanks.

### DIFF
--- a/lib/swoosh/adapters/smtp.ex
+++ b/lib/swoosh/adapters/smtp.ex
@@ -18,13 +18,16 @@ if Code.ensure_loaded?(:gen_smtp_client) do
           relay: "smtp.avengers.com",
           username: "tonystark",
           password: "ilovepepperpotts",
+          ssl: true,
           tls: :always,
           auth: :always,
           port: 1025,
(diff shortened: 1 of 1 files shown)


princemaple:
I'm not sure about adding it to the documentation, if gen_smtp doesn't mention it.
👍 on adding the option though. @stevedomin what do you think? FYI, the relevant code has been there for at least 7 years. (I couldn't be bothered to blame any further)

stevedomin:
I'm ok with having it in the docs, it's a well-known option of gen_smtp, despite being undocumented.

stevedomin:
Thanks @sergioaugrod!!

PR #210, opened 2018-03-12
Fix text only body

### LINKED ISSUE #207: SparkPost Adapter not working with text-only
```
web_1    |  {422,
web_1    |   %{
web_1    |     "errors" => [
web_1    |       %{
web_1    |         "code" => "1400",
web_1    |         "description" => "content.inline_images requires content.html",
web_1    |         "message" => "required field is missing"
web_1    |       }
web_1    |     ]
web_1    |   }}}
```
while sending a pure text email

princemaple:
it's currently never allowed because of the existence of inline_images key
changing it to conditionally adding attachment keys resolves the problem

fix #207


stevedomin:
Thanks @princemaple, good catch!

PR #214, opened 2018-03-22
Poison to Jason Migration

Waasi:
- Removed Poison and added Jason dependency
- Modified all adapters to use Jason instead of Poison
- Modified AdapterCase to use Jason instead of Poison as json decoder
- Modified Mailgun & Sparkpost test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 erl_crash.dump
 *.ez
 .elixir_ls/
+*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@
 erl_crash.dump
 *.ez
 .elixir_ls/
-*.swp
--- a/lib/swoosh/adapters/dyn.ex
+++ b/lib/swoosh/adapters/dyn.ex
@@ -45,7 +45,7 @@ defmodule Swoosh.Adapters.Dyn do
 
     case :hackney.post(url, headers, body, [:with_body]) do
       {:ok, 200, _headers, body} ->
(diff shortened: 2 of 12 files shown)


princemaple:
Hi @Waasi , thank you so much for picking up this one. PR looks good. One thing though, could poison be unlocked now? I don't think any of the dependencies directly depend on Poison? 👍 otherwise a great one, thanks again!

Waasi:
Hi again @princemaple, It seems that inch_ex depends on Poison and Plug depends on Poison through inch_ex. Thanks again.

princemaple:
Thank you for checking! I'll leave the merge to @stevedomin, as I don't have access to a work computer. Thanks again for the contribution! :blush:

.gitignore
```diff
@@ -5,3 +5,4 @@
 erl_crash.dump
 *.ez
 .elixir_ls/
+*.swp
```
stevedomin: Can you remove this line please? Keen to keep this file Elixir/Erlang specific.

yordis:
@stevedomin I would definitely make this a configuration because Jason has it own limitations (everything has to be Jason derive) for accomplish it performance.

Maybe we should add `config :swoosh, json_library: Jason` and fallback to Poison (because backwards compatibility) so every side is happy with it.

princemaple:
Switching to Jason in Swoosh won't be a breaking change. Everyone who
depends on poison in their own code still gets to use poison.

princemaple:
Thanks @Waasi !
I'll make the follow up change to make it configurable.

PR #216, opened 2018-03-27
Make json library configurable

princemaple:
Runs mix format in #217 to fix the long lines and brackets.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,2 +1,3 @@
 use Mix.Config
 
+config :swoosh, :json_library, Jason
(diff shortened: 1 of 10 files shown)


yordis:
@stevedomin @princemaple I would suggest to add the default JSON library on `@json_library Application.get_env(:swoosh, :json_library, Jason)` because it will no default library which I wouldn't mind of you put one by default.

yordis:
@princemaple that works when you are developing the module but no when you add Swoosh as a deps.

Unless I am completely wrong about it, but `config` works at the higher  application level only.

princemaple:
You are right.

PR #217, opened 2018-03-27
Run mix format

princemaple:
merge #216 first

manually excluded some changes, probably better refactor the code instead of formatting it poorly.


stevedomin:
I'd probably

stevedomin:
I'd probably format all the code even if it yields poor formatting, we can update it afterwards.

👍

princemaple:
Will do a second pass on the rest of the files at some point 👍

PR #219, opened 2018-03-28
Use xref exclude to suppress warnings instead of conditionally define modules

### LINKED ISSUE #202: Use xref exclude to suppress undefined module warning and emit explicit warning in validate_config
https://github.com/swoosh/swoosh/pull/187#issuecomment-363884842

This way it's probably clearer for users.

cc @stevedomin

princemaple:
close #202 

not doing the emit warnings in `validate_config` part, decided not to because you have no way to know if the module is actually used or not. if it's not used, any reminder is just noise.

Before this PR, when calling Mailer.deliver with SMTP adapter but missing `:gen_smtp`

```
** (UndefinedFunctionError) function Swoosh.Adapters.SMTP.validate_config/1 is undefined (module Swoosh.Adapte
rs.SMTP is not available)
    Swoosh.Adapters.SMTP.validate_config([otp_app: :sw, adapter: Swoosh.Adapters.SMTP, relay: "smtp.sparkpostm
ail.com", port: nil, tls: nil, auth: nil, username: "SMTP_Injection", password: "7f25919d44fca84d00f9eec82a668
8b35cb304b1"])
    (swoosh) lib/swoosh/mailer.ex:102: Swoosh.Mailer.deliver/2
```

After this PR

```
** (UndefinedFunctionError) function :mimemail.encode/2 is undefined (module :mimemail is not available)
    :mimemail.encode({"text", "html", [{"Content-Type", "text/html; charset=\"utf-8\""}, {"From", "test@test.com"}, {"To", "test@gmail.com"}, {"Subject", "test"}, {"Mime-Version", "1.0"}], [], "<h1>test<
/h1>"}, [])
    (swoosh) lib/swoosh/adapters/smtp.ex:45: Swoosh.Adapters.SMTP.deliver/2
```

bonus: warnings are gone if plug is not there. (we do have proper logging if plug was expected to be there)


stevedomin:
Do we have no way at all of knowing if a particular adapter is used? In `Swoosh.Mailer` for instance? If we do we could show a more explicit error.

princemaple:
@stevedomin please take a look at https://github.com/swoosh/swoosh/pull/219/commits/d43fef58e4d6d397dd455c0a6b3c59b6e4d9b524

This adds compile time warning, but takes quite some code.
```
> iex.bat -S mix
Compiling 1 file (.ex)

06:02:24.569 [error] The following dependencies are required to use Swoosh.Adapters.SMTP:

- gen_smtp_client from :gen_smtp
```

princemaple:
ping @stevedomin https://github.com/swoosh/swoosh/pull/219#issuecomment-377357159
could use some sanity checking before I move any further

princemaple:
Did some investigation on the test failure. It's caused by Elixir 1.4 or before [matching against {:module, _}](https://github.com/elixir-lang/elixir/blob/v1.4/lib/elixir/src/elixir_compiler.erl#L192) when trying to load compile stuff. 🙄

princemaple:
@stevedomin ready for review.

stevedomin:
@princemaple does this PR include changes from the formatter?

princemaple:
@stevedomin only the mix file was formatted. The rest is just removing the `if Code.ensure_loaded?` thingy. use this link to review without white space changes https://github.com/swoosh/swoosh/pull/219/files?w=1

stevedomin:
Looks good to me. Sorry for the wait on the review!

princemaple:
Are you OK with this part as well?

```elixir
+  if Version.match?(System.version(), "~> 1.5") do
+    # TODO: Remove version guard when dropping Elixir 1.4 support
+    # Elixir < 1.5 raises an unexpected error when on_load fails
+    test "raise if mailer defined with nonexistent adapter", %{valid_email: email} do
```

PR #223, opened 2018-03-29
Fix recompilation (potential) issue

yordis:
Because module attributes requires to recompile the deps to be
able to pick up the change. It is better to not use it for
configuration.

If somebody compiles the swoosh package with Jason and later on
decides to change it to Poison it will require to recompile the
swoosh deps.

Elixir do not recompile the deps unless it detects a change and
since change the config value is not change it will still have
the old value of the configuration.

### DIFF
--- a/lib/swoosh.ex
+++ b/lib/swoosh.ex
@@ -6,8 +6,6 @@ defmodule Swoosh do
   @doc false
   def version, do: @version
 
-  @json_library Application.get_env(:swoosh, :json_library, Jason)
-
   @doc false
-  def json_library, do: @json_library
+  def json_library, do: Application.get_env(:swoosh, :json_library, Jason)
 end


princemaple:
If they removed Jason without recompiling the dependencies, they would get a runtime error. I'd rather leave this to the developers than having to look this config up every single time. Besides, I don't think there is a reason to switch back to Poison, since Jason is API compatible and faster.

yordis:
@princemaple they will no get a runtime error because `Jason` is available in the project since it is a dependency of swoosh.

princemaple:
Then they will be all good. I still don't see a problem.

yordis:
What do you mean? If I want to configure another JSON library as you allowed me to, I am expecting to use it.

“Unless” you are aware of the compilation you wouldn’t expect that you need to force to recompile the swoosh dependency. 

Because from code perspective nothing change you would need to force the recompilation. While lead to more issues As I had been fixing in other packages because I am used to check that right away at this point.

princemaple:
I'm starting to think that there is no point letting people choose which json lib to use. In Swoosh, Jason or Poison,  it's virtually the same. As you said, Jason is now a dependency for Swoosh, people are going to have it anyway if they use Swoosh, or Ecto, or Phoenix. It's basically implementation detail in our case.

While I'm still considering whether this config is worth existing, whoever want to use Poison can force a recompilation. If they don't, Swoosh still works. I'm not worried. And having to get runtime config again and again for this now-I-think-less-useful config is what I'm trying to avoid.

princemaple:
The only way this would make sense is if we had both jason and poison as optional dependencies. However, we didn't have poison optional back then, and naturally not optional for jason either now. And we are probably not going to make them so because this would be a breaking change that gives little to no benefit.

princemaple:
I'm aware of the issue. That's why I did this https://github.com/swoosh/swoosh/pull/146 for the things that being configurable really provides value.

Thanks for sharing anyway!

PR #229, opened 2018-04-18
Add regex email body test assertions

michallepicki:
While using swoosh I wanted to assert with a regex that a particular string was being sent in the html_body. Currently the only way I could make it work is:

```
    assert_received({:email, email})
    assert email.html_body =~ regex
```

While it kind of works it doesn't read so well so I'm proposing adding regexp support
so we will be able to do things like:

```
assert_email_sent(to: {"Bob", "example@example.com"}, text_body: ~r/Hi Bob!/)
```
edit: I updated the PR so it no longer has three new options and pattern matches the value on regex struct type instead

### DIFF
--- a/lib/swoosh/test_assertions.ex
+++ b/lib/swoosh/test_assertions.ex
@@ -63,6 +63,14 @@ defmodule Swoosh.TestAssertions do
   defp assert_equal(email, {:text_body, value}), do: assert(email.text_body == value)
   defp assert_equal(email, {:html_body, value}), do: assert(email.html_body == value)
 
+  defp assert_equal(email, {:regex, regex}) do
+    assert_equal(email, {:regex_html, regex})
+    assert_equal(email, {:regex_text, regex})
+  end
+
+  defp assert_equal(email, {:regex_html, regex}), do: assert(email.html_body =~ regex)
+  defp assert_equal(email, {:regex_text, regex}), do: assert(email.text_body =~ regex)
+
   @doc ~S"""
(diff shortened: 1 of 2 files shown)


michallepicki:
We might want to rename `assert_equal` to something else now

lib/swoosh/test_assertions.ex
```diff
@@ -63,6 +63,14 @@ defmodule Swoosh.TestAssertions do
   defp assert_equal(email, {:text_body, value}), do: assert(email.text_body == value)
   defp assert_equal(email, {:html_body, value}), do: assert(email.html_body == value)
 
+  defp assert_equal(email, {:regex, regex}) do
+    assert_equal(email, {:regex_html, regex})
+    assert_equal(email, {:regex_text, regex})
```
princemaple: You may not have both of them present

princemaple: Might be a better idea to extend the previous clause and simply match regex `value` and do regex compare.

princemaple:
Thanks for the PR!

lib/swoosh/test_assertions.ex
michallepicki: Done

princemaple:
Added some tests for the new functionality. I'll merge once CI is passed.

princemaple:
Thank you for the contribution! @michallepicki

stevedomin:
Thanks @michallepicki, great addition!

PR #232, opened 2018-04-27
Added support for template_alias in Postmark adapter

sebastianseilund:
Thanks for Swoosh!

I'd like to use Postmark's [TemplateAlias](https://postmarkapp.com/developer/api/templates-api#email-with-template) feature in order to use the same alias across different Postmark servers for different Mix environments (one server for dev, one server for prod for example).


princemaple:
@stevedomin this added integration test requires some change on the testing postmark account. master is failing because of it.

PR #234, opened 2018-04-27
Display provider_options in mailbox_viewer

sebastianseilund:
Also made sure that _some_ content renders for Subject. Without it, the `.col-sm-10` element will have zero height and the whole layout is broken.

This is how it looks now:

screen shot 2018-04-27 at 12 23 08 pm


stevedomin:
👍

stevedomin:
@sebastianseilund we will need a rebase on this one as well. Thanks!

sebastianseilund:
Done

princemaple:
🎉

PR #240, opened 2018-05-18
Traivs

princemaple:
This change cuts down the build times by 1 minute or so.


stevedomin:
Nice one @princemaple !

PR #245, opened 2018-06-05
Expose Attachments in Mailbox Preview

maennchen:
screen shot 2018-06-05 at 17 23 48

### DIFF
--- a/lib/plug/mailbox_preview.ex
+++ b/lib/plug/mailbox_preview.ex
@@ -53,6 +53,35 @@ if Code.ensure_loaded?(Plug) do
       |> send_resp(200, email.html_body)
     end
 
+    get "/:id/attachments/:index" do
+      index = String.to_integer(index)
+      id
+      |> conn.assigns.storage_driver.get()
+      |> case do
+        %{attachments: attachments} when length(attachments) > index ->
+          attachments
+          |> Enum.at(index)
+          |> case do
+            %{data: data, content_type: content_type} when !is_nil(data) ->
+              conn
+              |> put_resp_content_type(content_type)
+              |> send_resp(200, data)
+            %{path: path, content_type: content_type} when !is_nil(path) ->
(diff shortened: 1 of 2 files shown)


lib/plug/mailbox_preview.ex
```diff
@@ -53,6 +53,35 @@ if Code.ensure_loaded?(Plug) do
...
+    get "/:id/attachments/:index" do
+      index = String.to_integer(index)
+      id
+      |> conn.assigns.storage_driver.get()
+      |> case do
+        %{attachments: attachments} when length(attachments) > index ->
+          attachments
+          |> Enum.at(index)
+          |> case do
+            %{data: data, content_type: content_type} ->
```
princemaple: This will probably always match? There is no nil guard here and the struct always has all the keys.

maennchen: You’ probably right, I’m going to fix that.

maennchen: Done

lib/plug/mailbox_preview.ex
```diff
@@ -53,6 +53,35 @@ if Code.ensure_loaded?(Plug) do
...
+    get "/:id/attachments/:index" do
+      index = String.to_integer(index)
+      id
+      |> conn.assigns.storage_driver.get()
+      |> case do
+        %{attachments: attachments} when length(attachments) > index ->
+          attachments
+          |> Enum.at(index)
+          |> case do
+            %{data: data, content_type: content_type} when not is_nil(data) ->
```
maennchen: Sorry about than one, I'm a little bit in a hurry today.

princemaple: No worries!

princemaple:
Thank you again for the PR!

princemaple:
@maennchen released in v0.15

Issue #247, opened 2018-06-12
Clearing message queue in mailbox preview


justanotherdot:
Is there a way to clear out the mailbox preview for local dev? I would find this useful after long sessions of debugging notifications and don't want to be lost in other emails being sent. 

It looks like `Swoosh.Adapters.Local.Storage.Memory` has a `:delete_all` call implemented. One option I can imagine working for this type of feature.

1. A route to invoke `:delete_all` in `mailbox_preview.ex`
2. A button to simplify invoking the above route

Would this be useful to others? Maybe I'm one of the few that might want this feature.

stevedomin:
We should definitely add a button. I've wanted that myself for a while.

Do you think you can you submit a PR for it?

justanotherdot:
Sure thing 👍

princemaple:
Thanks @justanotherdot !

PR #249, opened 2018-06-18
Delete queued messages with button in preview

justanotherdot:
Ref #247.

Sorry this took a second to get up! Works been keeping me busy.

The placement and styling of the button can probably be configured from here; what's the general thoughts?

Here's what it looks like as is:

kapture 2018-06-18 at 13 43 47

### DIFF
--- a/lib/plug/mailbox_preview.ex
+++ b/lib/plug/mailbox_preview.ex
@@ -46,6 +46,14 @@ if Code.ensure_loaded?(Plug) do
       |> send_resp(200, template(emails: emails, email: nil, conn: conn))
     end
 
+    get "/clear" do
+      conn.assigns.storage_driver.delete_all()
+
+      conn
+      |> put_resp_header("location", "/")
+      |> send_resp(302, '')
+    end
+
     get "/:id/html" do
       email = conn.assigns.storage_driver.get(id)
       conn
--- a/lib/plug/mailbox_preview.ex
+++ b/lib/plug/mailbox_preview.ex
@@ -39,19 +39,19 @@ if Code.ensure_loaded?(Plug) do
     plug :match
     plug :dispatch
 
-    get "/" do
-      emails = conn.assigns.storage_driver.all()
-      conn
-      |> put_resp_content_type("text/html")
-      |> send_resp(200, template(emails: emails, email: nil, conn: conn))
-    end
-
-    get "/clear" do
+    post "/clear" do
       conn.assigns.storage_driver.delete_all()
 
       conn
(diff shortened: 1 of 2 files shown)


princemaple:
Hi @justanotherdot Thanks for the PR!

We don't normally modify server data with a get request. Even though this is dev only, could we do that with a POST request? An inline form that wraps the button should do it. What do you think?

justanotherdot:
Yes, it is a little unprincipled. I don't mind either way; I can have the handler for the button produce a fetch and and redirect after the promise is resolved so as to choose a different method for the request.

lib/plug/templates/mailbox_viewer/index.html.eex
```diff
@@ -98,6 +98,7 @@
                 <h6 class="list-group-item-heading">Empty mailbox...</h6>
               </div>
             <% else %>
+              <button class="btn btn-primary btn-block" onclick="location.href = '/clear'">Delete Message Queue</button>
```
princemaple:
```html
<form method="POST" action="/clear">
  <button class="btn btn-primary btn-block" type="submit">Delete Message Queue</button>
</form>
```

lib/plug/mailbox_preview.ex
```diff
@@ -46,6 +46,14 @@ if Code.ensure_loaded?(Plug) do
       |> send_resp(200, template(emails: emails, email: nil, conn: conn))
     end
 
+    get "/clear" do
```
princemaple: `post "/clear"`

justanotherdot:
> something like this should do? no need for js.

Definitely, just a super rough idea that popped off the top of my head. Thanks for the reminder about form method attr's 🙂

I'll doll that up now.

justanotherdot:
@princemaple any thoughts on the layout for the button?

princemaple:
I haven't tried to run this yet, what does it look like now?

justanotherdot:
> I haven't tried to run this yet, what does it look like now?

Should be same as before, here's a new gif:

kapture 2018-06-18 at 16 26 01

The `margin-bottom` is to counteract the 1em `margin-bottom` being applied to `form`. I think that's possibly coming from bootstrap.

justanotherdot:
> The margin-bottom is to counteract the 1em margin-bottom being applied to form. I think that's possibly coming from bootstrap.

to make things a little bit more tidy: shall I move that up into the inline CSS up top?

princemaple:
That would be great, thanks. Thanks a lot!

PR #256, opened 2018-07-14
Add support for Mailgun tags

DavidOliver:
See https://documentation.mailgun.com/en/latest/user_manual.html#tagging


princemaple:
Code looks good.

> A single message may be marked with up to 3 tags.

Should we have a check on this, possibly throw a warning?
@stevedomin or do we just let the user to figure it out?

stevedomin:
I'm fine with leaving it as is. If we had checks I'd rather have them in an helper function but we don't really do this at all at the moment (provider-specific helper functions).

PR #260, opened 2018-07-25
Add Cowboy2

hl:
I'd like to add cowboy 2 to the mix.

The current version (2.4) still has `Plug.Adapters.Cowboy` and the mail preview works.


princemaple:
Thanks. Could you include the mix.lock change as well?

stevedomin:
@princemaple do you think we should do that yet?

princemaple:
`{:cowboy, "~> 1.0.1 or ~> 1.1 or ~> 2.4", optional: true},` I don't see anything harmful. Users can still choose which version to use..?

stevedomin:
True! 👍

princemaple:
Just tried locally, you are right.

PR #266, opened 2018-08-02
Decode SendGrid adapter 4xx response

jackmarchant:
👋 Thanks for the library, it's very well put together.

In the [SendGrid API](https://sendgrid.com/docs/API_Reference/Web_API_v3/How_To_Use_The_Web_API_v3/errors.html#-Failed-Requests) it provides error messages for fields, or even general messages but this is not decoded.
This pull request adds decoding for 4xx responses, so that when using the Sendgrid adapter, you can easily use the body from the response.


princemaple:
Thanks for the PR. I made a minor change to make the status code boundaries more consistent. I'll merge on CI pass.

princemaple:
@jackmarchant are you waiting on this to be released?

jackmarchant:
No I’m not waiting for it. Just saw it when looking through

PR #271, opened 2018-08-07
fix clear button path when base_path is set

jhchen:
Thanks for this library! The latest version added a clear button but it will not work if you have `base_path` set as we do in our use. This PR takes this option into account and constructs the urls accordingly.


princemaple:
Hi @jhchen , could you check if it still works with the commit I added? Thanks!

jhchen:
Yup still works for me! Thanks didn't know if I should change the default base_path but this makes it cleaner.

princemaple:
Thanks for the confirmation!

princemaple:
It's out in v0.16.1

PR #276, opened 2018-08-15
sendgrid - handlebar substitutions with dynamic_template_data

gnimona:
## Intention
- according to [sendgrid api doc](https://dynamic-templates.api-docs.io/3.0/mail-send-with-dynamic-transactional-templates/v3-mail-send), for handlebar substitutions, request body should contain `dynamic_template_data` inside `personalizations` object


princemaple:
Looks good, thanks. On phone now, will take another look later.

princemaple:
Thanks @gnimona

PR #277, opened 2018-08-15
Improve TestAssertions

### LINKED ISSUE #274: Impossible to do partial matching with `Swoosh.TestAssertions.assert_email_not_sent/1`
`Swoosh.TestAssertions.assert_email_not_sent/1` pins the `email` value and passes it to `refute_received`. Since pattern matching against pinned values compares the actual values instead of doing the pattern match (it's essentially a `==` comparison), it's impossible to use `assert_email_not_sent` with a partial email (like it's possible with `assert_email_sent`). For example:

```elixir
email = Swoosh.Email.new(to: "foo@bar.com")
Swoosh.Adapters.Test.deliver(email, [])

email.to() # => [{"", "foo@bar.com"}]

assert_email_sent(to: "foo@bar.com") # pass
assert_email_not_sent(email) # fail

assert %{to: [{"", "foo@bar.com"}]} = email # pass

assert_email_not_sent(%{to: [{"", "foo@bar.com"}]}) # pass (should fail if actually doing pattern matching)
```

It's possible to allow actual pattern matching if `assert_email_not_sent` is a macro:

```elixir
defmacro assert_email_not_sent(arg) do
  quote do
    refute_received {:email, unquote(arg)}
  end
end
```

Also, shouldn't it be named `refute_email_sent`?

13k:
This fixes #274.

Changes:

* added `assert_email_sent/0` that asserts any email was sent
* added `refute_email_sent/0` (macro) that asserts no email was sent ("alias" of `assert_no_email_sent/0`)
* added `refute_email_sent/1` (macro)

Besides implementing the simple pattern matching with `refute_email_sent(pattern)` - which requires exact matching on inner data structures - I took the opportunity to also add a `refute_email_sent(attributes)` version that creates a pattern (a Map) that automatically converts single values to single element lists while formatting the addresses (`to: "email@example.com"` becomes `to: [{"", "email@example.com"}]`).

Some notes:

* `mix format` changed several lines
* added `@spec`s
* added a stricter pattern matching on `assert_email_not_sent/1` (since it pins the `email` variable, it only works with actual `Email` structs - any other value can be a false positive - so it seems clearer if it accepts only `Email` structs)
* added more test cases for `assert_no_email_sent/0`

### DIFF
--- a/lib/swoosh/test_assertions.ex
+++ b/lib/swoosh/test_assertions.ex
@@ -78,14 +78,11 @@ defmodule Swoosh.TestAssertions do
   @doc ~S"""
   Asserts no emails were sent.
   """
-  @spec refute_email_sent() :: false | no_return
   defmacro refute_email_sent() do
     quote do
       refute_received {:email, _}
     end
   end
-
-  @spec refute_email_sent(keyword | Email.t() | map) :: false | no_return
 
   @doc ~S"""
   Asserts email with `attributes` was not sent.
--- a/lib/swoosh/test_assertions.ex
+++ b/lib/swoosh/test_assertions.ex
@@ -13,6 +13,16 @@ defmodule Swoosh.TestAssertions do
   alias Swoosh.Email
 
   @doc ~S"""
+  Asserts any email was sent.
+  """
+  @spec assert_email_sent() :: true | no_return
+  def assert_email_sent do
+    assert_received {:email, _}
+  end
+
+  @spec assert_email_sent(Email.t() | map) :: true | no_return
+
+  @doc ~S"""
   Asserts `email` was sent.
 
   You pass a keyword list to match on specific params.
@@ -37,7 +47,7 @@ defmodule Swoosh.TestAssertions do
 
   def assert_email_sent(params) when is_list(params) do
     assert_received {:email, email}
-    Enum.each(params, fn param -> assert_equal(email, param) end)
+    Enum.each(params, &assert_equal(email, &1))
   end
 
   defp assert_equal(email, {:subject, value}), do: assert(email.subject == value)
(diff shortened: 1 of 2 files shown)


test/swoosh/test_assertions_test.exs
```diff
@@ -4,87 +4,107 @@ defmodule Swoosh.TestAssertionsTest do
...
-        """
-        No message matching {:email, ^email} after 0ms.
-        The following variables were pinned:
-          email = #{inspect(wrong_email)}
-        Process mailbox:
-          {:email, #{inspect(email)}}
-        """
-      )
+      String.trim("""
+      No message matching {:email, ^email} after 0ms.
```
princemaple: is this change from formatter?

13k: yep, I have my editor to auto format on save with `mix format`, all formatting changes are from `mix format`.

lib/swoosh/test_assertions.ex
```diff
@@ -66,16 +76,76 @@ defmodule Swoosh.TestAssertions do
   defp assert_equal(email, {:html_body, value}), do: assert(email.html_body == value)
 
   @doc ~S"""
-  Asserts `email` was not sent.
+  Asserts no emails were sent.
+  """
+  @spec refute_email_sent() :: false | no_return
```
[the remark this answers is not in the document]
13k: Good question, I'm not sure one should document the meta types (at compile time) or the runtime types. I'm not even sure if it's correct to `@spec` macros.

After digging a little deeper, reading elixir's source code, it seems macros shouldn't be `@spec`ed. [`is_nil/1` (macro)](https://github.com/elixir-lang/elixir/blob/2b338092b6da5cd5101072dfdd627cfbb49e4736/lib/elixir/lib/kernel.ex#L2577-L2593) doesn't have a spec and [`is_number/1` (function)](https://github.com/elixir-lang/elixir/blob/2b338092b6da5cd5101072dfdd627cfbb49e4736/lib/elixir/lib/kernel.ex#L547-L556) does.

That's my bad to submit before making sure.

lib/swoosh/test_assertions.ex
```diff
@@ -124,13 +124,7 @@ defmodule Swoosh.TestAssertions do
...
 
   defp email_pattern(%{} = pattern, {key, value}) when key in [:to, :cc, :bcc] do
-    value =
-      case value do
-        addresses when is_list(addresses) -> value
-        address -> [address]
-      end
-
-    Map.put(pattern, key, Enum.map(value, &format_recipient/1))
+    Map.put(pattern, key, value |> List.wrap() |> Enum.map(&format_recipient/1))
```
13k: Thanks for that. I searched for something to wrap a single value in `Enum` and didn't find. Forgot to check `List`.

princemaple:
Didn't know that `refute_x` actually return false when success. Learned something new.

princemaple:
Thanks again @13k !

13k:
No problem, thanks for the work on swoosh :)

[`refute_receive` specifically](https://github.com/elixir-lang/elixir/blob/v1.7.2/lib/ex_unit/lib/ex_unit/assertions.ex#L877) returns `false`. I'm not sure about others. Looks like `refute` returns whatever the expression evaluated to.

PR #287, opened 2018-09-11
handlebar substitutions with dynamic_template_data

steffenix:


princemaple:
Thanks @steffenix , could you rebase to the latest master (github isn't offering to merge master into this branch, maybe there is a big diff?) and provide a link to the documentation of the added feature?

princemaple:
On this page https://sendgrid.com/docs/API_Reference/Web_API_v3/Mail/index.html I don't see anything about `dynamic_template_data`. I think the feature you are trying to implement is already there? `substitutions`

steffenix:
I will see with @chakming if he can update is branch. We were looking at https://sendgrid.com/docs/ui/sending-email/how-to-send-an-email-with-dynamic-transactional-templates/

princemaple:
What's the difference between `substitutions` and `dynamic_template_data`?

steffenix:
we tried substitutions and it was not working with  transactional templates.

princemaple:
Yeah it's so weird. One version of their docs I found uses `substitutions` and I also found another version that uses `dynamic_template_data`. Since I don't use SendGrid myself, I have no idea which one is correct, or if both are needed for different situations.

steffenix:
We were confused too.

princemaple:
https://github.com/swoosh/swoosh/pull/276 isn't this commit already merged in this PR?

Issue #286, opened 2018-09-10
Add sandbox mode for sendgrid


rhnonose:
I would like to have the option to send the e-mail in a sandbox mode, more specifically for sendgrid.

[It's through a field on the payload](https://sendgrid.com/docs/for-developers/sending-email/sandbox-mode/) , but other providers such mailgun and mandrill have a test mode (different api keys).

It could be added as an adapter specific config or as a parameter in the `Email` struct (couldn't find other providers that operate this way but it might be the case).

princemaple:
We have `provider_options` that can be used for provider specific data/operations.
https://github.com/swoosh/swoosh/blob/master/lib/swoosh/email.ex#L32

A few examples:
- https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/mandrill.ex#L165
- https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/sendgrid.ex#L117

You can send a PR to enable sandbox mode for sendgrid adapter.

rhnonose:
Thanks @princemaple, I'll work on it.

princemaple:
Great, thank you!

PR #291, opened 2018-09-14
Add mail_settings to enable sandbox_mode for sendgrid adapter

rhnonose:
Implements #286


princemaple:
Looks good, thanks!

princemaple:
Would you be able to add some documentation for this feature too?

rhnonose:
Couldn't find a place to put it, should it be in `lib/swoosh/adapters/sendgrid.ex` or `lib/swoosh/email.ex`?

princemaple:
In the module doc of sendgrid.ex will be great!

Issue #258, opened 2018-07-17
Copy attachments for dev mailbox


DanielDent:
The development mode mock adapter is extremely useful, but currently has a significant limitation: the attached file appears to simply be a pointer to wherever in the filesystem the attachment was located at the time the email was created.

Normal program operation might involve that attachment being deleted immediately after the email is sent, or it might involve a file whose contents change over time. 

Copying the attachment data into a new file would make the dev mock adapter considerably more useful for workflows which involve attachments.

princemaple:
Hi Daniel,

Thanks for reporting!

We could, in this function https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/local/storage/memory.ex#L119-L123
call https://github.com/swoosh/swoosh/blob/master/lib/swoosh/attachment.ex#L78-L80 with :base64
to store the attachments as either provider options or private fields.

Do you feel like you can send a PR?

princemaple:
Closing with #306

PR #306, opened 2018-10-24
Store attachment content in Memory when using local adapter

theodowling:
I've been playing with the Test adapter and found this issue #258.

This should resolve it, by embedding the raw data of the attachment in the `Genserver` "storage".
I didn't convert to `:base64` to keep compatible with the rendering side where it doesn't decode: https://github.com/swoosh/swoosh/blob/e3beba436797850ceb664ceee8d5178487a82617/lib/plug/mailbox_preview.ex#L73-L76

Please let me know if you need any changes made.


princemaple:
Hi @theodowling, I'm currently traveling (and sick). If Steve doesn't get to it first, I'll take care of this in a few days. Thanks for the PR!

princemaple:
Thanks @theodowling !

Issue #309, opened 2018-11-08
Is there a way to set timeout time?


rparcus:
It would be useful to be able to control timeout time. How can we do that?

stevedomin:
Hello @rparcus,

What do you mean exactly? Which timeout are you referring to?

stevedomin:
We don't support this currently, however this would be nice to have.

princemaple:
Should be resolved by #327

PR #310, opened 2018-11-12
Pass request options to hackney

Doerge:
Adds the ability to pass options to the underlying hackney request.

It would make sense to consider a generic way of handling this across all adapters.

Might be a solution to #309 as well.


Doerge:
Would be useful if we could set it in the config as well:
```
  adapter: Swoosh.Adapters.Mandrill,
  api_key: "SG.x.x"
  request_options: [{:connect_timeout, 15_000}, {:recv_timeout, 15_000}]
```

princemaple:
Sorry, thought the other PR #311 is also sent by you.

We are going to hold back on merging the two PRs, because we'd prefer to add an abstraction that works on all adapters, instead of patching every adapter with the same code.

Doerge:
A generic way would be way better. Feel free to close this PR then :)

PR #311, opened 2018-11-15
✨ Pass hackney request options on sendgrid

### LINKED ISSUE #309: Is there a way to set timeout time?

rparcus:
Based on #310 and comments, this allows for controlling hackney directly while using Sendgrid.
Closes #309 for me.


princemaple:
close in favor of #327

PR #326, opened 2018-12-26
Temporarily skip integration tests

princemaple:
Revert this later when integration tests are fixed


princemaple:
FYI @stevedomin

Issue #340, opened 2019-02-05
SMTP adapter + html results in long lines in quoted-printable message.


ivan-kolmychek:
We are trying to switch our emails to Swoosh from Rails-based system.
We use service to try to predict are our mails counted as spam or not, that includes running them through Barracuda spamfilter.

When we switched, this started to show up in report:
> MIME_QP_LONG_LINE_2 - META: RAW: Quoted-printable line longer than 76 chars

Looking at the quoted-printable source of result, the lines do exceed 76 chars.

As an attempt to quickly fix it, I have locally cloned swoosh and switched encoding of parts to base64 and it looks like that solves the problem.

Does anyone else has this problem?
Would PR with setting to change encodings between quoted-printable and base64 be of any interest to anyone?

princemaple:
released v0.22.0

PR #341, opened 2019-02-05
Add option to change encoding to smtp adapter

ivan-kolmychek:
Ref #340.

This was born as a "quick and dirty" fix, so it may not be perfect. 
Tests are passing, but I have not added any new ones. 
It also does not solve the original issue itself, just side-steps it with using different encoding.

Feedback and advice on how to improve it (and what tests to add) are welcome.


ivan-kolmychek:
@princemaple thanks. Of course, take your time. 

We are using this in the meantime as fork, so maybe we'll spot any other nuances as well.

princemaple:
Hi @ivan-kolmychek , how's your change been going for you? I made a few modifications. I'll merge this one in if there hasn't been any issue on your side.

ivan-kolmychek:
@princemaple it's running all right, we have not found any issues so far.

I pulled your changes and tested it a bit locally now, it does works for me.

After you will merge I think we can switch to using `master` for now, so it will get more live testing until you are ready to actually release next version.

princemaple:
Thanks a lot for testing this live :)

princemaple:
Hi @ivan-kolmychek , just checking, has master caused any trouble so far?

ivan-kolmychek:
@princemaple no, have not spotted any problems yet, but we pinned it to the [merge commit](https://github.com/swoosh/swoosh/commit/4f2196ca73e061329b1e493f9d12dc5aeb1c5f07) for now, so I can't say anything about any further changes.
We can swap it to current HEAD on Monday (Feb 18th) to test that through a bit more.

princemaple:
Thanks for the reply. The later commits should not have an impact on the SMTP code. I think I will make a release sometime next week. Thanks again!

PR #342, opened 2019-02-05
Gmail adapter

mpran:
Gmail adapter to send email as the authorized user. Requires `access_token` to be passed in for every request instead of hardcoded config api key.

Because google only accepts `message/rfc822` content-type, had to implement parser for it so the tests are not failing. Open to discussion if that seems like a problem.

Also wasn't too sure about outside dependency, `Mail` is needed to properly render the message, but let me know if I need to fix anything.

Thanks


lib/swoosh/adapters/gmail.ex
```diff
@@ -0,0 +1,145 @@
...
+    |> prepare_to(email)
+    |> prepare_cc(email)
+    |> prepare_bcc(email)
+    |> prepare_subject(email)
+    |> prepare_text(email)
+    |> prepare_html(email)
+    |> prepare_attachments(email)
+    |> prepare_reply_to(email)
+    |> Mail.Renderers.RFC2822.render()
+    |> parse_bcc(email)
```
mpran: worth noting(maybe should document also) that this is on purpose, because `Mail` will remove any `bcc` header when rendering, so we have to put it back after to be able to use the functionality. Wasn't able to dig out the exact reason for it, but if anyone has any info on that would love to know.

princemaple:
Hi @mpran, thanks for the PR and the detailed explanation! Just letting you know that I have seen this one, but will take a couple of weeks before I will have time to review it. Been busy lately, and this one is gonna take me some research.

princemaple:
FYI @stevedomin

lib/swoosh/adapters/gmail.ex
princemaple: Yep, a comment will be super helpful in the future, in case my OCD kicks in. Also this function should probably be private.

princemaple:
> Requires `access_token` to be passed in for every request instead of hardcoded config api key.

Stupid question: how come we can't config it? does it change all the time? OAuth access token shouldn't really change, should it? Unless scope changes. - I'm probably missing something simple..

> Also wasn't too sure about outside dependency, `Mail` is needed to properly render the message, but let me know if I need to fix anything.

Optional dependencies are fine. We don't want to DIY everything :)

Not a complete review! Just had a few minutes so thought I could kick this off first.

lib/swoosh/adapters/gmail.ex
mpran: Ah good catch. It was private, but changed it to access directly while doing some testing. Will update and add the comment

mpran:
> > Requires `access_token` to be passed in for every request instead of hardcoded config api key.
> 
> Stupid question: how come we can't config it? does it change all the time? OAuth access token shouldn't really change, should it? Unless scope changes. - I'm probably missing something simple..
> 

When app is authorized for the first time, you will get a `refresh_token` and `access_token` ([more info](https://developers.google.com/identity/protocols/OAuth2WebServer#offline)). Access token is only valid for 1 hour by default, but you can send `expires_in` with n seconds ([more info](https://developers.google.com/identity/protocols/OAuth2UserAgent#handlingtheresponse)) and when it expires, you need to use refresh token to request a new access token. The way we currently do this is capture expiration date and time of the access token and before sending an email, we do a quick check to make sure it's not expired, if it is, we just request a new one and update it. This works well for us, because emails are sent from a service when users are not present.

> > Also wasn't too sure about outside dependency, `Mail` is needed to properly render the message, but let me know if I need to fix anything.
> 
> Optional dependencies are fine. We don't want to DIY everything :)

Yeah, wouldn't be easy to render the message without the library :)
> 
> Not a complete review! Just had a few minutes so thought I could kick this off first.

Appreciate it!

lib/swoosh/adapters/gmail.ex
```diff
@@ -99,7 +99,10 @@ defmodule Swoosh.Adapters.Gmail do
     |> prepare_attachments(email)
     |> prepare_reply_to(email)
     |> Mail.Renderers.RFC2822.render()
-    |> parse_bcc(email)
+    # When message is rendered, bcc header will be removed and we need to prepend bcc list to the
+    # begining of the message. Gmail will handle it from there.
+    # https://github.com/DockYard/elixir-mail/blob/master/lib/mail/renderers/rfc_2822.ex#L161
```
princemaple:
```suggestion
    # https://github.com/DockYard/elixir-mail/blob/v0.2.0/lib/mail/renderers/rfc_2822.ex#L139
```

Master changes, link to master is not reliable. Also, too bad they haven't made a release since 2017, although they are doing things on the lib.

princemaple:
Hi @mpran , I've made a few not so important modifications. Please take a look and let me know if everything still looks alright! Thanks a lot for your patience and sorry about the wait.

mpran:
@princemaple thanks for taking a look and making those changes! Everything looks good to me. I will submit another PR with readme update if this is good to go.

princemaple:
@mpran thanks a lot for another great PR :)

PR #346, opened 2019-02-10
Allow preview server to start when using plug cowboy version 2

aselder:
`Plug.Adapters.Cowboy` has been deprecated and replaces with the `plug_cowboy` package.

If `plug_cowboy` version 2.0.0 or higher is installed, the preview server will not start because `child_spec/4` has been removed.

This changes the optional dependency to use the `plug_cowboy` package (which will indirectly require plug), and changes to using the `child_spec/1` function which is available in both versions 1 and 2 of `plug_cowboy`


princemaple:
Thank you very much for the PR @aselder !

PR #349, opened 2019-02-10
skip inch doc on PR builds

princemaple:


princemaple:
does not seem to work

PR #351, opened 2019-02-12
Rename test support files to exs

princemaple:


princemaple:
Nope needs to be ex, or explicitly loaded?

PR #359, opened 2019-02-27
Sparkpost Adapter: Attach attachment correctly to the json body

LostKobrakai:
`attachments` and `inline_images` are part of the content sent in the body and not at the root: https://developers.sparkpost.com/api/transmissions/#transmissions-post-send-inline-content


princemaple:
Thanks a lot @LostKobrakai . Going through the blame history looks like this was introduced in one of my commits that was meant to fix another bug. 🤕 apologies if that caused much inconvenience.

PR #365, opened 2019-03-13
Update smtp.ex

### LINKED ISSUE #364: Not enforcing password and username to be string with auth: :never
We upgraded to v0.22 and got an issue:

When using `Swoosh.Adapters.SMTP` with  `auth: :never` the username and password need to be a string and cant be `nil` because they are checked by `enforce_type!` which fails with:

> ** (ArgumentError) password is not configured properly, got: nil, expected a string

This seemed to introduce that change: https://github.com/swoosh/swoosh/pull/166

princemaple:
https://github.com/Vagabond/gen_smtp/blob/master/src/gen_smtp_client.erl#L363

Indeed `:auth` can be `:never`.

close #364 

@h4cc would simply removing the username and password in the config work for you as a solution to the other half of the issue?


h4cc:
Thanks, will test this.

PR #259, opened 2018-07-18
Mailjet adapter

denispeplin:
Not 100% ready, some preparations are disabled, but already usable.

I would leave it in open pull requests in case someone needs it.

PR #367, opened 2019-03-14
Mailjet

princemaple:
Close #259


princemaple:
@panmichal thanks a lot for taking this one forward. I'll review it as a whole soon.

panmichal:
Hmm, okay, test should use `Swoosh.AdapterCase` instead of `AdapterCase`, looks like it has changed on master.

princemaple:
Thanks. That's correct.

princemaple:
@panmichal @denispeplin Thanks a lot for your contributions!

augnustin:
Hey guys!

Great to read that the Mailjet adapter is here! :smile: 

The doc hasn't been updated though, this could certainly help.

But at first, is it possible print out a usage example here? It may not be difficult but I'm slightly confused, especially regarding where to set the template_id!

Cheers!

panmichal:
@augnustin you can use `put_provider_option(:template_id, your_template_id)`.

Good catch with the docs, I'll try to fix it.

PR #372, opened 2019-03-21
add sent_at when sending locally

### LINKED ISSUE #371: Sent time in mailbox
It would be nice if the local development mailbox would show a datetime of when the selected email was sent.

princemaple:
close #371

### DIFF
--- a/lib/plug/templates/mailbox_viewer/index.html.eex
+++ b/lib/plug/templates/mailbox_viewer/index.html.eex
@@ -138,7 +138,7 @@
                   <dd class="col-sm-10"><%= render_recipient(@email.to) %></dd>
 
                   <dt class="col-sm-2">Sent at</dt>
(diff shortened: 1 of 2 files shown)


princemaple:
@LostKobrakai what do you think about this? need some js to convert it to locale string (local timezone)?

LostKobrakai:
Looks good. For local time it could probably use https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString to keep it simple.

LostKobrakai:
Maybe something like that:
```
  <time datetime="2019-03-21T13:15:22.500704Z">2019-03-21T13:15:22.500704Z</time>
  …
  <script>
    var times = document.getElementsByTagName("time");
    for(var i = 0; i < times.length; i++){
      var t = times[i];
      var date = new Date(t.getAttribute("datetime"));
      t.innerHTML = date.toLocaleString();
    }
  </script>
```

Issue #314, opened 2018-11-20
Support for Erlang 21 in the SMTP Adapter


paulruescher:
Came across an issue with the SMTP Adapter and Erlang 21:

```
an exception was raised:
    ** (MatchError) no match of right hand side value: '0A'
        (gen_smtp) .../deps/gen_smtp/src/mimemail.erl:898: :mimemail.encode_quoted_printable/3
```

We found that by updating `{:gen_smtp, "~> 0.12.0"}` to `{:gen_smtp, "~> 0.13.0"}` in our `mix.exs` file fixes our problem.

Just wanted to raise this in case there should be an actual update within this project.

ilyakard:
I've just had the same problem, it took up a bit of debugging... The readme files for both Swoosh and the SES adapter still reference gen_smtp 0.12 instead of 0.13

princemaple:
Hi,

Sorry about the inconvenience.
I'm currently traveling. Could you send a PR?

PR #382, opened 2019-04-24
bump gen_smtp version to work with erlang 21, update readme

ilyakard:
#314

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ end
     def deps do
       [
         {:swoosh, "~> 0.23"},
-        {:gen_smtp, "~> 0.12"}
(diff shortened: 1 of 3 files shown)


princemaple:
Thanks a lot!

ilyakard:
No worries, thanks for your quick reply! I haven’t had time to write a test unfortunately, would require erlang 21, gen_smtp 0.12 and a “\n” in the message body to trigger the error (I think).

princemaple:
It's hard to test with different versions of deps.
I'll try and make a release (to update the docs) when I get better internet.
Thanks again.

PR #388, opened 2019-06-07
Add documentation for :preview_port

dweremeichik:
Changing the preview port is not mentioned in any of the swoosh documentation here or on hex. This took me a while to dig up. Hopefully it will save others time in the future.


princemaple:
Thanks @dweremeichik

PR #392, opened 2019-07-08
add 'perform_deliveries' option to disable send an emails

parallel588:


parallel588:
hi all.
 what do you think  could we add 'perform_deliveries' option to disable send an emails ?

princemaple:
Hi @parallel588 could you give a reason to add this? This looks to me more like something you do on your application level. If you want to disable sending, simply wrap the deliver function from Swoosh with a function of your own, and in your own function, take config/predict into consideration. There is no reason Swoosh shouldn't be sending an email. In local env we can still send via the local adapter to an in-memory inbox.

parallel588:
hi @panmichal , I do it for pleroma.
https://git.pleroma.social/pleroma/pleroma/issues/1062
https://git.pleroma.social/pleroma/pleroma/merge_requests/1389

we will use this feature to dynamic enable\disable send emails (in admin panel).

I seen similar option in ActionMailer (rails world)

princemaple:
Wrong @ but fine

In this case I think my previous comment applies quite well. I'm going to close this PR. Swoosh should not be handling this kind of business logic. Instead, you should write your own module that wraps and replaces normal deliver calls.

Rails world is convenient but has a lot of things done wrong.

parallel588:
ok, thaks.

PR #400, opened 2019-08-24
README: Improve email preview documentation when using phoenix.

donileo:


princemaple:
Excellent update, thanks for the PR!

PR #402, opened 2019-08-30
Local

### LINKED ISSUE #401: Crashing when set in distributed multi-node environment
We have a 15 node cluster where swoosh is available on a multitude of the nodes, however when one node crashes or there's a few going down and coming up, other nodes that we're testing with (that are literally doing nothing but idling) get crashed by swoosh and taken down which in turn crashes the entire node. 

```elixir
[info] global: Name conflict terminating {Swoosh.Adapters.Local.Storage.Memory, #PID<42923.3171.0>}
[info] global: Name conflict terminating {Swoosh.Adapters.Local.Storage.Memory, #PID<42922.3174.0>}
[info] Application swoosh exited: shutdown
{"Kernel pid terminated",application_controller,"{application_terminated,swoosh,shutdown}"}
Kernel pid terminated (application_controller) ({application_terminated,swoosh,shutdown})
```

We're not using swoosh in any way on the other nodes so I'm going to remove the dependency from there, but never the less this is strange behaviour I wanted to report. 

This is probably related to #333

princemaple:
@MikaAK please let me know if this is clear

close #401


MikaAK:
Super clear we can just disable it and that'll be perfect 🙏  thanks for the fast action @princemaple

PR #406, opened 2019-10-07
Add mailjet error delivery config

augnustin:
Add `TemplateErrorDeliver` and `TemplateErrorReporting` `provider_options` to mailjet

### DIFF
--- a/lib/swoosh/adapters/mailjet.ex
+++ b/lib/swoosh/adapters/mailjet.ex
@@ -79,7 +79,7 @@ defmodule Swoosh.Adapters.Mailjet do
     |> prepare_reply_to(email)
     |> prepare_attachments(email)
     |> prepare_variables(email)
-    |> prepare_template_id(email)
+    |> prepare_template(email)
     |> prepare_custom_headers(email)
     |> wrap_into_messages
     |> Swoosh.json_library.encode!()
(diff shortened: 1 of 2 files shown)


lib/swoosh/adapters/mailjet.ex
```diff
@@ -139,10 +139,12 @@ defmodule Swoosh.Adapters.Mailjet do
 
   defp prepare_variables(body, _email), do: body
 
-  defp prepare_template_id(body, %{provider_options: %{template_id: template_id}}) do
+  defp prepare_template_id(body, %{provider_options: %{template_id: template_id} = provider_options}) do
     body
     |> Map.put("TemplateID", template_id)
     |> Map.put("TemplateLanguage", true)
+    |> Map.put("TemplateErrorDeliver", provider_options.template_error_deliver)
+    |> Map.put("TemplateErrorReporting", provider_options.template_error_reporting)
```
princemaple: They will raise if the keys don't exist. Consider using `[:template_error_deliver]`. Will the provider behave OK when these 2 keys are provided but are nil?

augnustin:
My bad, I'm still confused between

```
kw.key
kw[:key]
kw["key"]
```

which ones are optional/mandatory and set at runtime/set at compile time, but this is definitely `kw[:key]` required here.

I updated the PR.

Regarding the provider's behavior, [the docs](https://dev.mailjet.com/guides/#templates-error-management) doesn't seem to consider the field as being mandatory and since it was not there previously, it should behave as expected. :smile:

princemaple:
OK the tests are still failing. There are a few things we need to update.

1. the changed function is no longer only about the template id, we might want to change it to `prepare_template`
1. the new fields are added unconditionally, it's fine, we just need to add them to the test
1. I read the docs, the added fields are boolean, so it's probably ok to have the keys always present. However, it might be a good idea to double negate the values, so they are always boolean, and are `false` instead of `nil` when not provided.
```elixir
!!map[:key_that_may_not_exist] # => gives false when key missing
```

princemaple:
Hi @augnustin before I merge this in, just want to confirm with you that you have tested it with the provider, as I don't use mailjet I have no idea if the implementation is correct.

princemaple:
The way to test it is simply using swoosh from your local version. https://hexdocs.pm/mix/Mix.Tasks.Deps.html

augnustin:
Sure, I'm currently testing this and will let you know if anything goes wrong, but I reckon there are few chances this doesn't work. :smile:

augnustin:
@princemaple :+1: I can confirm it works for me.

PR #408, opened 2019-10-16
Added :custom_id mailjet API option.

augnustin:
I forgot to add the `CustomID` API option which we use. 

Doc is here: https://dev.mailjet.com/guides/#send-an-email-with-a-custom-id

I took the chance to make a little refactoring in the mailjet test file.


princemaple:
Cheers. I'm traveling at the moment and will have to make the release later. You can use the lib via github dependency for now.

princemaple:
@augnustin thanks for the PR, it's out in 0.24

augnustin:
Thanks for merging. I think we're good with Mailjet API integration for now. :smile:

PR #414, opened 2019-10-22
add mailgun template support

### LINKED ISSUE #412: Mailgun support for templates
To support mailgun templating, we should add `template`

Reference:
https://documentation.mailgun.com/en/latest/api-sending.html#sending

steffenix:
closes #412


princemaple:
Thanks for covering custom variables as well.

PR #418, opened 2019-10-25
encode map and list only

### LINKED ISSUE #417: mailgun, string variables should not be jason encoded
When using `prepare_custom_vars` a simple string should not be json encoded.

steffenix:
closes #417


princemaple:
Just to confirm, do numbers, booleans and null work without encoding?

princemaple:
Thanks for bringing it up and sending the PR :)

steffenix:
I will try.

princemaple:
@steffenix thanks a lot.

steffenix:
boolean, numbers and null all are ok.

princemaple:
Fantastic, cheers.

PR #439, opened 2020-02-13
Update .travis.yml

princemaple:


grahamhay:
How do I leave this Hotel California of a mailing list. Unsubscribe just does not do it!

princemaple:
@grahamhay you probably want to unwatch, or switch to watch releases only

princemaple:
ProTip: you most likely don't want to **watch** a repo that you don't develop on, stick to :star: :)

Issue #436, opened 2020-02-03
Feature: batch-send of messages (Mailjet)


panmichal:
Some APIs for, for example Mailjet, support sending multiple messages with one request.

https://dev.mailjet.com/email/guides/send-api-v31/#send-a-basic-email - here, `Messages` is an array and the adapter just wraps single email struct in a list to conform with the API.
I could implement handling multiple messages but would like to know your opinion about how it should be handled:
- handling list argument in `delivery/2` https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/mailjet.ex#L34
- separate function - `deliver_many` or something similar

princemaple:
I vote for a separate function, so we can keep the current signature and have clear logic path. Please add an optional callback, deliver_many or batch_deliver, to Adapter and implement it in Mailjet Adapter.

panmichal:
Makes sense! 
I'll work on a PR then.

princemaple:
Thanks! :star2:

princemaple:
I forgot to make a release didn't I?

princemaple:
It's out in 0.25

princemaple:
Thanks for the contribution!

PR #440, opened 2020-02-27
deliver_many/2 callback and implementation for Mailjet

panmichal:
As discussed in https://github.com/swoosh/swoosh/issues/436

Return values for `deliver/2` are left unchanged, for both success response and errors.
In `deliver/2` messages was a one-element list and I've made it a bit more generic so most of the logic is shared between those 2 functions.

### DIFF
--- a/lib/swoosh/adapter.ex
+++ b/lib/swoosh/adapter.ex
@@ -27,12 +27,16 @@ defmodule Swoosh.Adapter do
   @doc """
   Delivers an email with the given config.
   """
-  @callback deliver(email, config) :: {:ok, term} | {:error, term}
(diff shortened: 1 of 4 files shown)


princemaple:
Thanks for the PR! Will review it tomorrow or over the weekend.

princemaple:
Hi @panmichal please take a look at my changes. Thanks.

FYI there seems to be some change with Travis CI and I couldn't restart the build. I've re-pushed it in #442 . If everything is good I might have to close this and merge #442.

panmichal:
Hi @princemaple looks better now :) And still works, so 👍

princemaple:
closing in favor of #442

PR #447, opened 2020-03-28
Add ability to set SparkPost transmission options

florish:
This adds the ability to set `options` as defined in the SparkPost transmissions API documentation:

https://developers.sparkpost.com/api/transmissions/#header-request-body

Note that the `options` value is passed as-is to the JSON body sent to SparkPost.

I could not find any test for the SparkPost adapter, but in case I missed something and additional tests are needed, please let me know.


princemaple:
Hi @florish , just letting you know I've seen this PR. Recent travis integration change has caused me some issue maintaining this project. I'm waiting for some help from the original owner. Thanks for the contribution!

PR #450, opened 2020-04-02
Fix v0.25.1 release date in CHANGELOG.md

florish:
@princemaple Thanks for releasing a new version so quickly, really helpful!

Just spotted a copy-paste error in CHANGELOG.md, this PR is a fix for that.


princemaple:
:) Thank you sir!

PR #451, opened 2020-04-09
Ensure local storage adapter is started if enabled

wojtekmach:
Before this patch it may not have been started e.g. in an umbrella:
say you correctly configured with `local: true`, but because a domain
app may not depend on plug & cowboy, we returned empty children instead
of the local adapter (if configured)


princemaple:
Thanks!

Issue #455, opened 2020-04-22
Return email body from adapters


sadir:
Hi,

I have a use case where we use `Swoosh` to send an email via `Sendgrid`, but then we want to upload the fact that email has been sent into our CRM. They have an API to upload email activity.

This was fine before we used this library because the response from Sendgrid's API has the email that was sent in the `body`. We had built our own `Sendgrid.HTTPClient` and took the details of the email sent from that.`Swoosh` however, does not return the response body.

See: https://github.com/swoosh/swoosh/blob/v0.25.2/lib/swoosh/adapters/sendgrid.ex#L45

I don't want to recreate the email content myself because that's what the dynamic templates are for.

Is there any reason why the Sendgrid adapter is dropping the response body and only returning the message id? Would you consider returning the response body as well?

I've spotted that a decision was made originally to just return the id here: https://github.com/swoosh/swoosh/issues/52

princemaple:
That'd be before I started maintaining this repo. I think we can return more information. Essentially whatever the provider returns. Luckily we are returning a map, we can add the response to it without breaking the current API. Do you feel like putting up a PR?

sadir:
Yeah of course - I'll probably get to it next week at some point but 👍

princemaple:
Looking forward to that. Thanks!

sadir:
Draft PR up to fix this, early feedback welcome.

PR #456, opened 2020-04-24
Return email body from Sendgrid adapter when emails are sent successfully

sadir:
Fixes https://github.com/swoosh/swoosh/issues/455.

Going to turn the raw response body into a map, let me know now if you don't want to do that?

Also, I've got two other commits that you may or may not want. 

We use [asdf](https://github.com/asdf-vm/asdf) to manage versions per-project at Nested, thought you may want to consider using that to make development onboarding easier.

We mix format every file at Nested to remove debate about styling, spaces and indentation, etc. You may want to consider using that to make development onboarding easier.

Entirely up to you on these! I'm happy to accept your guidance.


sadir:
Looks like you have the [Plug.Parsers.JSON](https://hexdocs.pm/plug/Plug.Parsers.JSON.html) in use, so maybe the test data is wrong? It should already be a map?

princemaple:
Hi @sadir thanks for the quick PR. I'm in AEST so it's about mid night now. I'll review this and get back to you over the weekend or next week. Stay safe and have a good one!

princemaple:
Hi @sadir , running the formatter over the files you worked with is great. I did format many files along the way. I just didn't do it over the whole project. A separate PR to do that is welcome as well!

`.tool-version` I'm not sure. I use Windows and asdf doesn't work here. Also, I'm always on latest Elixir (1.10+) and OTP (22.3 now). I think it may not be wanted for everyone. I think it's more useful in company projects than in libraries.

Regarding the tests, you might want to replace `respond_with` with something like `Plug.Conn.resp(conn, 200, success_response)`, that might make a difference.

I'm OK with parsing response body, as long as the service provider documents that they return json response on successful requests. Please make a note in the adapter moduledoc as well.

Thanks again for the PR!

sadir:
Right - should all be there for you now. I've made it ready for review!

PR #468, opened 2020-05-21
Define an ApiClient behaviour and make Hackney an optional dep

josevalim:


lib/swoosh.ex
```diff
@@ -6,8 +6,6 @@ defmodule Swoosh do
   @doc false
   def version, do: @version
 
-  @json_library Application.get_env(:swoosh, :json_library, Jason)
-
   @doc false
-  def json_library, do: @json_library
+  def json_library, do: Application.fetch_env!(:swoosh, :json_library)
```
josevalim, line 10: Moved this to runtime for consistency.

mix.exs
```diff
@@ -45,17 +46,21 @@ defmodule Swoosh.Mixfile do
   end
 
   def application do
-    [extra_applications: [:logger], mod: {Swoosh.Application, []}]
+    [
+      extra_applications: [:logger],
+      mod: {Swoosh.Application, []},
+      env: [json_library: Jason, api_client: Swoosh.ApiClient.Hackney]
```
princemaple, line 52: TIL `env: [...]`

PR #469, opened 2020-05-22
Fix warnings on Elixir master

josevalim:
Apologies, I introduced a warning. I used the opportunity to fix other warnings too.


princemaple:
Thank you!

PR #471, opened 2020-05-25
Add syntax highlighting to attachments example

mbuhot:
While reading the README I noticed the attachments section wasn't syntax highlighted.
This PR adds a code fence around it.


princemaple:
Thank you @mhuhot :)

PR #480, opened 2020-06-25
Fix behaviour of test adapter

narrowtux:
The test adapter's deliver implementation should still return the sent email like the other adapters do.


princemaple:
The adapters don't actually return the original email. Neither does the type spec say it should return the email.

princemaple:
Let me know if you have strong opinions that this should be the way.

narrowtux:
No, I probably misunderstood then. At least one of the adapters does return the email like this so I thought this was an inconsistency.

PR #483, opened 2020-06-29
fixed a return result for Sendmail

parallel588:


princemaple:
Thanks @parallel588 .

princemaple:
I'm not quite sure about a new release with this fix. This is technically a bug fix, but also a breaking change. Might take some time to polish a few things then possibly make a 1.0 release. However I'm quite busy at the moment. Apologies if this takes a bit longer to go out than normal. 😢

princemaple:
#484 FYI

PR #490, opened 2020-07-14
Remove misleading Swoosh.Adapters.Gmail clause about omitting `from` while building email

### LINKED ISSUE #489: Avoiding setting `from` address as Gmail set it for me
Hi.

As [Swoosh.Adapters.Gmail docs](https://hexdocs.pm/swoosh/Swoosh.Adapters.Gmail.html#content) says, I don't need to set `from` and it would be ignored if it wouldn't match account's one.

So this one returns an error:
```ex
%Swoosh.Email{}
|> to({"Me", "my@gmail.com"})
|> subject("Hi")
|> html_body("<div>Hi</div>")
|> ProlingWeb.Mailer.deliver
# {:error, :from_not_set}
```

My umbrella app config:
```ex
config :proling_web, ProlingWeb.Mailer,
  adapter: Swoosh.Adapters.Gmail,
  access_token: {:system, "GMAIL_API_ACCESS_TOKEN"}
```

When I adding a `from` pipe it works:
```ex
%Swoosh.Email{}
|> to({"Me", "my@gmail.com"})
|> from("No reply", "noreply@proling.ru")
# ... the rest
```

What I'm doing wrong or maybe I should fix docs or code in some point?

vadimshvetsov:
Closes #489

Also adding `access_token` existence to config looks helpful

### DIFF
--- a/lib/swoosh/adapters/gmail.ex
+++ b/lib/swoosh/adapters/gmail.ex
@@ -18,6 +18,10 @@ defmodule Swoosh.Adapters.Gmail do
         adapter: Swoosh.Adapters.Gmail,
         access_token: {:system, GMAIL_API_ACCESS_TOKEN}
 
+      # To deal with token refresh, it could be a better idea to pass the access token
+      # in via deliver config explicitly, if you don't update the environment variable
+      # periodically. e.g.
+      MyMailer.deliver(my_email, access_token: my_access_token)
 
       # lib/sample/mailer.ex
       defmodule Sample.Mailer do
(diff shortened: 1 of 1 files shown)


vadimshvetsov:
@princemaple your addition looks very helpful, I've faced a problem with renewing token. What is the proper way to set it for the first time and then renew it and save accross deploys? Documenting this would be very helpful because after I've faced restrictions from Gmail API with 3600 seconds expiration I started think about migrating to SMTP adapter.

princemaple:
It's very easy to implement a GenServer that requests a token on start, then refresh every half hour or so, whatever the rate limit is. Every deliver call just needs to get the token from the GenServer.

PR #518, opened 2020-10-05
Add support for temporary credentials in AmazonSES

riiwo:
In case you use IAM role for authenticating AWS requests you can fetch
temporary access_key and secret_key from that role, but you also need to
include additional X-Amz-Security-Token header to that request.

Docs: [Http request signature](https://docs.aws.amazon.com/general/latest/gr/sigv4-add-signature-to-request.html)

### DIFF
--- a/lib/swoosh/adapters/amazon_ses.ex
+++ b/lib/swoosh/adapters/amazon_ses.ex
@@ -159,6 +159,7 @@ defmodule Swoosh.Adapters.AmazonSES do
     |> prepare_header_date(current_date_time)
     |> prepare_header_length(query)
     |> prepare_header_authorization(query, current_date_time, config)
+    |> prepare_header_security_token(config)
     |> Map.to_list()
   end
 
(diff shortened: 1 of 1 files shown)


lib/swoosh/adapters/amazon_ses.ex
```diff
@@ -231,6 +232,13 @@ defmodule Swoosh.Adapters.AmazonSES do
     }"
   end
 
+  defp prepare_header_security_token(headers, config) do
+    case config[:security_token] do
+      nil -> headers
+      _ -> Map.put(headers, "X-Amz-Security-Token", config[:security_token])
```
princemaple, line 238:
```suggestion
     token -> Map.put(headers, "X-Amz-Security-Token", token)
```

princemaple:
Travis isn't doing its job, I don't know why. I'll have to ask the owner of this repo to check.
/cc @stevedomin

princemaple:
Actually I intend to revert this merge. This can be achieved by `email |> header(""X-Amz-Security-Token", my_token)`. Could you verify? @riiwo 

Thanks.

princemaple:
cc maybe @indrekj

indrekj:
@riiwo maybe you can test it?

From reading the code, I don't think that would work. Hackney doing the post request using ApiClient takes headers as a second argument https://github.com/swoosh/swoosh/blob/master/lib/swoosh/api_client.ex#L85 . These are passed in from https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/amazon_ses.ex#L76 which do not read `email` headers. In this case the request headers and email headers are different things.

princemaple:
I think you are correct @indrekj . I didn't read carefully enough.
However, could we change it to be a [provider option](https://hexdocs.pm/swoosh/Swoosh.Email.html?#put_provider_option/3), since it's AWS related, instead of a random config passed in?
Also, some documentation update would be greatly appreciated!

indrekj:
Using the provider option does make sense (I didn't see that option before).

PR #519, opened 2020-10-05
Add support for custom Content-Type header

riiwo:
Amazon SES uses SMTP helper that adds Content-Type header for all the email.

In case Content-Type header is already added, sending email will fail as SES
return error for multiple Content-Type headers.
```
%{code: "InvalidParameterValue", message: "Duplicate header 'Content-Type'."}
```


princemaple:
Replaced with #522

PR #532, opened 2020-10-12
Mailjet: short circuit deliver_many to prevent error when delivering 0 emails

Adzz:
Sometimes I'll do this:

```elixir
some_list
|> Enum.map(&to_swoosh_email)
|> Swoosh.Adapters.Mailjet.deliver_many()
```

and sometimes some_list will be an empty list. Currently if it is empty
I will get an error from mailjet, when really it isn't an error, I just
have no emails to send.

This short circuits the email send when there are no emails to send.

Is it a good idea?


princemaple:
I'd go with an empty list instead of a random string.

Adzz:
Good idea, shall I also implement similar for the other adapters ?

princemaple:
IIRC only mailjet and test adapter implement deliver many at the moment. If you could take a look at test adapter, and maybe implement it for local adapter, it'd be greatly appreciated. Feel free to just ignore those, though. :)

PR #541, opened 2020-10-29
Fix gen_smtp check

princemaple:


lib/swoosh/adapters/smtp/helpers.ex
```diff
@@ -18,9 +18,13 @@ defmodule Swoosh.Adapters.SMTP.Helpers do
     mime_encode(type, subtype, headers, parts, encoding_config)
   end
 
-  Application.load(:gen_smtp)
   gen_smtp_major =
-    :gen_smtp |> Application.spec(:vsn) |> to_string() |> Version.parse!() |> Map.get(:major)
+    if Code.ensure_loaded?(:gen_smtp) do
```
gmile, line 22: @princemaple this line always returns `false` for me. Should this be?: 

```elixir
Code.ensure_loaded?(:gen_smtp_application)
```

princemaple: Right.

PR #548, opened 2020-11-08
Utilize Enum.map_join/3

tmock12:
I noticed one place where the code is mapping over lists, then joining them (two iterations), so I did a search and found a few more. While they are small and the performance gain is probably negligible, any little bit helps? I don't think it's any more difficult to reason about and I think it's also good to promote Enum.map_join/3 as a "best practice". Thanks for Swoosh!


princemaple:
Thanks!

PR #549, opened 2020-11-12
Add support for arrays in mailgun custom_vars

rschef:
The current way in `Swoosh.Adapters.Mailgun` to send variables is to use a form parameter, which is not recomended as it’s limited to simple key value data. If you have arrays, dictionaries in values or complex json data you have to supply variables via X-Mailgun-Variables header, which is implemented on this PR.

https://documentation.mailgun.com/en/latest/user_manual.html#templates


princemaple:
Thanks @rschef . This looks good.

princemaple:
Out in v1.1.1

rschef:
Thank you @princemaple!

PR #554, opened 2020-12-08
remove alias Swoosh.Mailer from quote block

felipelincoln:
having
```elixir
alias Swoosh.Mailer
```
inside the `quote/2` in the `Swoosh.Mailer` module is bad.

I have an application called `Mailer` which have a bunch of submodules, and if I 
```elixir
use Swoosh.Mailer, otp_app: :mailer
```
I can not alias any of my submodules afterwards
```elixir
alias Mailer.HTML
```
Because it is trying to get the `Swoosh.Mailer.HTML` instead of my `Mailer.HTML`

```
** (UndefinedFunctionError) function Swoosh.Mailer.HTML.render/3 is undefined (module Swoosh.Mailer.HTML is not available)
```

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -63,11 +63,10 @@ defmodule Swoosh.Mailer do
   """
 
   alias Swoosh.DeliveryError
+  alias Swoosh.Mailer
 
   defmacro __using__(opts) do
     quote bind_quoted: [opts: opts] do
-      alias Swoosh.Mailer
-
       @otp_app Keyword.fetch!(opts, :otp_app)
       @mailer_config opts
 
@@ -144,7 +143,7 @@ defmodule Swoosh.Mailer do
     Application.get_env(otp_app, mailer, [])
     |> Keyword.merge(mailer_config)
     |> Keyword.merge(dynamic_config)
-    |> Swoosh.Mailer.interpolate_env_vars()
+    |> Mailer.interpolate_env_vars()
   end
 
   @doc """
@@ -178,7 +177,7 @@ defmodule Swoosh.Mailer do
         :abort
 
       {:error, deps} when is_list(deps) ->
-        Logger.error(Swoosh.Mailer.missing_deps_message(adapter, deps))
+        Logger.error(Mailer.missing_deps_message(adapter, deps))
         :abort
     end
   end


princemaple:
Sorry this would be a breaking change. I cannot merge it in, at least not before 2.0

There is a simple workaround - place your `alias` above `use`:
```elixir
iex(1)> defmodule X.C do
...(1)>   defmacro __using__(_) do
...(1)>     quote do
...(1)>       alias X.C
...(1)>     end
...(1)>   end
...(1)> end
{:module, X.C,
 <<70, 79, 82, 49, 0, 0, 5, 52, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 145,  
   0, 0, 0, 14, 10, 69, 108, 105, 120, 105, 114, 46, 88, 46, 67, 8, 95, 95, 105,
   110, 102, 111, 95, 95, 10, 97, 116, ...>>, {:__using__, 1}}
iex(2)> defmodule C.D do
...(2)> end
{:module, C.D,
 <<70, 79, 82, 49, 0, 0, 3, 188, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 129, 
   0, 0, 0, 13, 10, 69, 108, 105, 120, 105, 114, 46, 67, 46, 68, 8, 95, 95, 105,
   110, 102, 111, 95, 95, 10, 97, 116, ...>>, nil}
iex(3)> defmodule M do
...(3)>   alias C.D
...(3)>   IO.inspect D
...(3)>   use X.C
...(3)>   IO.inspect C
...(3)> end
C.D
X.C
{:module, M,
 <<70, 79, 82, 49, 0, 0, 3, 184, 66, 69, 65, 77, 65, 116, 85, 56, 0, 0, 0, 127,
   0, 0, 0, 13, 8, 69, 108, 105, 120, 105, 114, 46, 77, 8, 95, 95, 105, 110,   
   102, 111, 95, 95, 10, 97, 116, 116, 114, ...>>, X.C}
```

felipelincoln:
It's ok! Yeah I'm currently using this workaround, thanks for the detailed comment by the way.  
Let me ask you, but what is the purpose of this alias inside the quote? I can't get it

princemaple:
There are a few `Mailer` calls below. I guess I threw it in 3 years ago just so that I didn't have to type `Swoosh.Mailer` on those lines. Now we look at it, it shouldn't be there. Back then, 🤷‍♂️ I didn't think this through.

felipelincoln:
Well @princemaple, you have a bunch of `Mailer` call as well as `DeliveryError` calls inside that `quote/2`, which you are aliasing outside. I can't see why this PR would break it by moving the alias out of the quote.

princemaple:
The fact that it was aliased inside the quote, means users could access the alias, which is also why caused the issue to you. And if users have written code that in the same module that utilized the alias, moving it outside the quote means user code would break because of the removal of the alias. It is a breaking change, albeit an easy fix.

felipelincoln:
Alright, I got you!

princemaple:
Thanks for understanding

PR #562, opened 2021-01-21
Fix Mailbox Viewer

thiamsantos:
I think in the [new version of bootstrap](https://github.com/swoosh/swoosh/commit/854fcc09a6a4892280e4bafe83468270704f0ba2) the offset class was changed, because of that the preview in the new version is breaking. 

References:

- https://getbootstrap.com/docs/4.6/layout/grid/#offsetting-columns


princemaple:
Thanks!

PR #563, opened 2021-01-21
Add json endpoint for mailbox preview

### LINKED ISSUE #61: Add json endpoint for mailbox preview
When doing end-to-end testing with frontend of features like account registration that involve unique tokens and things like that, it would be useful to have a json version of the mailbox that the tests could query.

I did recently something similar in one of our projects. If you find this idea worth considering I could work on it.

thiamsantos:
Closes #61 

As [pointed out in the issue](https://github.com/swoosh/swoosh/issues/61#issuecomment-638032058) we have two alternatives, add JSON for the existing endpoint or create a new one specific for the JSON payload. Initially I thought in adding a different endpoint may be simpler for the daily usage, as we will not have to remember to pass the accept header to get the expected payload, a different endpoint seems to be more explicit. However I'm open to change to the a single endpoint in case is considered better for swoosh.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,13 @@ If you are curious, this is how it looks:
 
 ![Plug.Swoosh.MailboxPreview](https://github.com/swoosh/swoosh/raw/main/images/mailbox-preview.png)
 
+
+The preview is also available as a JSON endpoint.
+
+```sh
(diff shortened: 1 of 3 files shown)


princemaple:
Not having to specify a header is a good point. Thanks!

Issue #537, opened 2020-10-26
Make `from` optional


tofran:
Hello, thank you for the great work.

# Context:

We are creating an adapter for [Sendingblue](https://www.sendinblue.com/).
In Sendingblue if you do not specify `from` in the API request, it will delegate this to whats configured in the template editor (on their platform).

# Problem:

Currently `lib/swoosh/mailer.ex` prevents empty `from` addresses:

```
def deliver(%Swoosh.Email{from: nil}, _config) do
    {:error, :from_not_set}
  end
  def deliver(%Swoosh.Email{from: {_name, address}}, _config)
      when address in ["", nil] do
    {:error, :from_not_set}
  end
```

And we get the error: `{:error, :from_not_set}`

# Question:

What is the best way to make `from` optional?
How would you do it for our use case? (Without ugly hacks)
If we want to contribute our adapter to this repo how can we integrate this feature in your codebase?

Many thanks in advance.

EDIT: improved wording

princemaple:
Instead of allowing empty `from`, I might expand the current implementation to accept atoms. And in your adapter code you could look for a specific atom that denotes the intention to use the configured value, and then ignore it.

tofran:
@princemaple this would involve a refactor in your API, right?
Meaning you would have to update existing adapters?

princemaple:
Not really. It would just be an addition. Existing adapters don't necessarily need to handle atom `from`. Most changes will be in email.ex, nailer.ex and format.ex

princemaple:
Let me know if you need any help.

princemaple:
Thanks for the contribution @tofran . The new adapter is released in v1.2.0. The task to remove enforced `from` field is tracked in #556 as a future breaking change.

tofran:
Very good point, Sendinblue actually rejects empty subject:

```json
{
  "code": "missing_parameter",
  "message": "subject is required"
}
```

So case closed, theres no problem with the current implementation

princemaple:
@tofran thanks for the confirmation!

Issue #556, opened 2020-12-09
Breaking TODOs


princemaple:
For v2, which is too far in the future

- [ ] Reduce unnecessary code injection like #554

josevalim:
The list looks good but, FWIW, I would skip renaming `Swoosh.Adapters.*`. It is a small change but likely to have a huge impact on users, so probably not worth it?

princemaple:
Yup. This doesn't have to happen.

caioaao:
why is ` Do not assume any field is required` considered a breaking change?

princemaple:
We require `from`. If we remove this check,  anyone who relies on this would lose this guarantee.

PR #564, opened 2021-01-21
Sendinblue: allow using template's subject

tofran:
Adds the possibility for sending an email without providing the subject, and therefore using the one specified in the template.

Same as what was requested on #537, although this time it does not error, it provides an empty struct because of:
```elixir
defmodule Swoosh.Email do
  # (...)
  defstruct subject: "",
```
`lib/swoosh/email.ex` 

Related to #537


princemaple:
This has nothing to do with #556 , subject has never been required.

tofran:
Thank you @princemaple.
Yes you are right its not related to said issue.

I didn’t go this route because now you are preventing the usage of an empty subject.
Although thats not very desirable in a real life scenario, its the default swoosh behaviour on all other adapters.
From my side doesn’t make any difference tho

princemaple:
If you don't set a template id, it'd still be empty subject, no?

lib/swoosh/adapters/sendinblue.ex
```diff
@@ -119,6 +119,8 @@ defmodule Swoosh.Adapters.Sendinblue do
     Map.put(payload, "bcc", Enum.map(bcc, &prepare_recipient/1))
   end
 
+  defp prepare_subject(payload, %{subject: ""}), do: payload
```
tofran, line 122: Maybe changing this to 
```elixir
defp prepare_subject(payload, %{subject: ""}), do: payload
```
would solve the inconstancy

tofran:
I think we are not doing that check - but should
Thats a good test scenario

princemaple:
I meant, does it make a difference, when you send `""` as subject to sendinblue, and when you not send a subject?

tofran:
*Somehow my comment is now showing, if this is a dupe sorry please ignore.

Yes thats a good point, looks like they actually don't allow `"subject": "",`:

```json
{
  "code": "missing_parameter",
  "message": "subject is required"
}
```

All ok, case closed

princemaple:
Thank you very much for the quick responses and the confirmation 🎉

PR #565, opened 2021-01-22
Support integration tests

### LINKED ISSUE #66: Add support for integration testing
The current TestAdapter breaks with integration testing  using hound or wallaby.

The best option seem to be implementing some kind of ownership-based mechanism similar to what Ecto 2 does (https://github.com/elixir-lang/ecto/issues/1237)

@barisbalic @hubertlepicki @lau I would love to get your thoughts on that

thiamsantos:
Closes  #66

The approach took is really similar to one [suggested in the issue](https://github.com/swoosh/swoosh/issues/66#issuecomment-387516542).
I also considered a allowance mechanism similar to what ecto and mox do, but IMO we can accomplish a similar result with a much simpler solution. For the use case of swoosh, just making sure that the messages are sent to the test process seems to be enough to allow users to do integration tests.

### DIFF
--- a/lib/swoosh/adapters/test.ex
+++ b/lib/swoosh/adapters/test.ex
@@ -40,6 +40,10 @@ defmodule Swoosh.Adapters.Test do
   # Essentially finds all of the processes that tried to send an email (in the test)
   # and sends an email to that process.
   defp pids do
-    Enum.uniq([self() | List.wrap(Process.get(:"$callers"))])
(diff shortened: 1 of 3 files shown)


princemaple:
Hi @thiamsantos, thanks for another great PR. I made a couple changes. Could you take a look?

thiamsantos:
@princemaple The documentation is much better now after your changes, thanks :)

princemaple:
Thanks for taking another look! I'll make a release next week.

PR #566, opened 2021-01-28
Link to pre-1.0/changelog.md instead of readme.md

woolfred:
Thanks for your work! Only a minor thing, but from the link title this is the intended linking, I guess? At least this is what I would have expected.


princemaple:
My bad :) Thanks!

PR #567, opened 2021-02-02
Support Sendgrid `tracking_settings` as per API docs

craigp:
Adds support for Sendgrid `tracking_options` as per docs at https://sendgrid.com/docs/api-reference/

```elixir
new() |> put_provider_option(:tracking_settings, %{subscription_tracking: %{enable: false}})
```


princemaple:
FYI I rewrote most of the provider option related stuff in https://github.com/swoosh/swoosh/commit/4312ce0154f6deb35e26c05ea86478c2665308aa as they are getting too much

princemaple:
Released in v1.3.1

princemaple:
Thanks for the PR!

PR #568, opened 2021-02-03
Update postmark.ex

matthewford:
You cannot use 'subject' when using a template.


princemaple:
Hi @matthewford , just wondering, does it cause an error when subject is specified when using a template?

matthewford:
```
{:error,
 {422,
  %{
    "ErrorCode" => 1123,
    "Message" => "Neither 'TextBody', 'HtmlBody', or 'Subject' may be specified when using a template for the email."
  }}}
```

princemaple:
Thanks!

PR #570, opened 2021-02-03
Remove base_path in favor of using script_name from Plug.Conn

### LINKED ISSUE #569: Use conn.script_name for MailboxPreview
Currently one needs to set `base_path` manually for the preview plug to create proper links. But `conn.script_name` does already hold the necessary information.

princemaple:
close #569 
cc @LostKobrakai 

I'm keeping the config because I haven't checked how it works under multiple forward. If you have more information, maybe we can remove it altogether.

### DIFF
--- a/lib/plug/mailbox_preview.ex
+++ b/lib/plug/mailbox_preview.ex
@@ -37,7 +37,7 @@ if Code.ensure_loaded?(Plug) do
     def call(conn, opts) do
       conn =
         conn
-        |> assign(:base_path, opts[:base_path] || Enum.join(["/" | conn.script_name]))
+        |> assign(:base_path, opts[:base_path] || Path.join(["/" | conn.script_name]))
         |> assign(:storage_driver, opts[:storage_driver] || Memory)
 
       super(conn, opts)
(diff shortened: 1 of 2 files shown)


lib/plug/mailbox_preview.ex
```diff
@@ -37,7 +37,7 @@ if Code.ensure_loaded?(Plug) do
     def call(conn, opts) do
       conn =
         conn
-        |> assign(:base_path, opts[:base_path] || "/")
+        |> assign(:base_path, opts[:base_path] || Enum.join(["/" | conn.script_name]))
```
LostKobrakai, line 40: Shouldn't this be `Path.join` instead of `Enum.join`?

princemaple: True, that's what went through my head but didn't make it into the code.
Code review wins :)

LostKobrakai:
> I'm keeping the config because I haven't checked how it works under multiple forward.

Afaik `script_name` is meant to accumulate stripped segments exactly so that one can rebuild the proper request path. This is e.g. also done by the phoenix router helpers.

princemaple:
Thanks @LostKobrakai !

PR #576, opened 2021-03-03
Allow passing an anonymous function to assert_email_sent

### LINKED ISSUE #573: Asserting on provider options in tests
We're sending emails like so:
```
defmodule SomeEmail do
  import Swoosh.Email

  def new(recipient, param) do
    new()
    |> to(recipient)
    |> put_provider_option(:template_alias, "a-template")
    |> put_provider_option(:template_model, %{
      param: param
    })
  end
end
```

There's currently no way to assert the provider options which means that we can't really assert on the content of the emails.

I think this could be solved either by:
- Adding matcher for `:provider_option` to [`assert_email_received`](https://github.com/swoosh/swoosh/blob/main/lib/swoosh/test_assertions.ex#L93), or
- Adding support for passing a function:
```
assert_email_received(fn email ->
  # ...
end)
```
The function could either run it's own `assert` or return a boolean.

stefanchrobot:
Closes #573.

The function allows setting expectation on any condition, most notably on provider options.

The function is not supported for `assert_email_not_sent` nor `refute_email_sent` because they translate to pattern matching inside a `receive do ... end` block.

### DIFF
--- a/lib/swoosh/test_assertions.ex
+++ b/lib/swoosh/test_assertions.ex
@@ -51,7 +51,6 @@ defmodule Swoosh.TestAssertions do
     end
   end
 
-
   @doc ~S"""
   Asserts any email was sent.
   """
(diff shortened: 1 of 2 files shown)


princemaple:
This looks good. Thanks!

princemaple:
I try my best to release ASAP :) plus, small releases are much better.

Issue #585, opened 2021-03-25
Set custom headers in Gmail adapter


reichert621:
Is it possible to set custom headers in the Gmail adapter?

I'm trying to set the `"References"` and `"In-Reply-To"` in order to send messages as replies to threads (https://developers.google.com/gmail/api/guides/sending). Am I approaching this wrong?

princemaple:
If you could leave a link to the docs or even send a PR, it'd be greatly appreciate. Cheers.

reichert621:
hey @princemaple! before I dig into it I just wanted to double check that I'm thinking about this correctly... I'm just trying to send a message as a reply to an existing thread. am I right in thinking I just need to figure out a way for these headers to be passed though?

princemaple:
Hi @reichert621 I'm not familiar with Gmail API, or how to send a reply to a thread. Maybe @mpran (who contributed the gmail adapter) can chime in?

princemaple:
If it's only a matter of setting headers, should be easy to find out 😄

reichert621:
I hope it's that simple! I'll wait for @mpran to chime in, but otherwise I can look into it tomorrow :+1:

reichert621:
@princemaple I gave it a shot in this PR, but so far no luck with the replies... does anything look off in the code to you? https://github.com/swoosh/swoosh/pull/587

princemaple:
There is a chance that the suffix of your subject needs to match the original subject. Just guessing...

princemaple:
Correctly setting the custom headers, but I can't get the reply to show up in a thread either.

reichert621:
yeah, i discovered that the gmail `thread_id` also needs to be passed through in the request in order for this to work 🤷

hindkaer:
Hey!

Im struggling with the exact same issue, but i can't seem to find out how i should pass the `thread_id` into the Swoosh Email object. Could you explain that for me? :)

PR #587, opened 2021-03-30
Support custom headers in the Gmail adapter

reichert621:
Trying to fix https://github.com/swoosh/swoosh/issues/585

So far this doesn't seem to work for replies, but I'm hoping this is the right track? (It does seem to add the headers to the message now at least.)


lib/swoosh/adapters/gmail.ex
```diff
@@ -143,4 +144,12 @@ defmodule Swoosh.Adapters.Gmail do
 
   defp prepare_reply_to(body, %{reply_to: nil}), do: body
   defp prepare_reply_to(body, %{reply_to: reply_to}), do: Mail.put_reply_to(body, reply_to)
+
+  defp prepare_custom_headers(body, %{headers: nil}), do: body
```
princemaple, line 148:
```suggestion
  defp prepare_custom_headers(body, %{headers: headers}) when headers == %{}, do: body
```

headers default to an empty map.

princemaple (changes requested):
Looks OK in general, could you also add a test? Cheers.

princemaple (approved):

PR #601, opened 2021-05-17
Add metadata support to Postmark adapter

jakubpawlowicz:
Postmark supports adding arbitrary metadata (basically a hash) to API email sending call, see https://postmarkapp.com/developer/api/email-api#send-a-single-email

This PR adds support for inserting that metadata into API call via generic `put_provider_option/2` call.

This is my first swoosh PR so please forgive me if I did anything wrong or re-implemented an already existing feature!


princemaple:
This looks good. Could you also add a section to the module doc and talk about the provider options? It'd be great if you could cover the existing ones but OK to only mention the one you added, I can add the rest when I get time.

princemaple:
I got some time so I just added them.

princemaple:
Released in 1.3.10, thanks!

jakubpawlowicz:
This was fast, thanks a lot @princemaple :100:

princemaple:
:)

PR #614, opened 2021-07-25
Adds telemetry to Mailer.deliver & Mailer.deliver_many

### LINKED ISSUE #613: Telemetry events
Hi,

Thanks for created this useful library.

Would you be interested in adding `telemetry` events?
I think it would be useful for folks that want to have a dashboard for to track their emails.

# Example

Attach a telemetry handler for sent email:

```elixir
# in Application.start
:telemetry.attach("my-email-handler", [:swoosh, :mail, :sent], &MyHandler.handler_event/4, nil)
```

Then in the handler, push the event to StatsD, DataDog, Appsignal etc..

```elixir
defmodule MyHandler do
  def handle_event([:swoosh, :mail, :sent], measurement, meta, _) do
     # send to statsd or wtvr dashboard 
  end
end
```

If it's something you're interested in, LMK, and I'll open a PR.

joshnuss:
This PR adds the following telemetry events:

- `[:swoosh, :deliver, :start]`: occurs when `Mailer.deliver/2` starts
- `[:swoosh, :deliver, :stop]`: occurs when `Mailer.deliver/2` succeeds
- `[:swoosh, :deliver, :exception]`: occurs when `Mailer.deliver/2` fails
- `[:swoosh, :deliver_many, :start]`: occurs when `Mailer.deliver_many/2` starts
- `[:swoosh, :deliver_many, :stop]`: occurs when `Mailer.deliver_many/2` succeeds
- `[:swoosh, :deliver_many, :exception]`: occurs when `Mailer.deliver_many/2` fails

## Usage

Example of capturing a count of emails that succeed and fail:

```elixir
defmodule MyHandler do
  def handle_event([:swoosh, :deliver, :stop], _measurements, metadata, _config) do
     StatsD.increment("mail.sent.success", 1, %{mailer: metadata.mailer})
  end

  def handle_event([:swoosh, :deliver, :exception], _measurements, metadata, _config) do
     StatsD.increment("mail.sent.failure", 1, %{mailer: metadata.mailer})
  end 

  def handle_event([:swoosh, :deliver_many, :stop], _measurements, metadata, _config) do
     StatsD.increment("mail.sent.success", length(metadata.emails), %{mailer: metadata.mailer})
  end

  def handle_event([:swoosh, :deliver_many, :exception], _measurements, metadata, _config) do
     StatsD.increment("mail.sent.failure", length(metadata.emails), %{mailer: metadata.mailer})
  end 
end

:telemetry.attach_many("my-handler", [
   [:swoosh, :deliver, :stop],
   [:swoosh, :deliver, :exception],
   [:swoosh, :deliver_many, :stop],
   [:swoosh, :deliver_many, :exception],
 ], &MyHandler.handle_event/4, nil)
``` 

## TODO

- [x] add info to readme

Closes #613

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -76,33 +76,33 @@ defmodule Swoosh.Mailer do
 
   You can capture events by calling `:telemetry.attach/4` or `:telemetry.attach_many/4`. Here's an example:
 
-  ```
-  # tracks the number of emails sent successfully/errored
-  defmodule MyHandler do
-    def handle_event([:swoosh, :deliver, :stop], _measurements, metadata, _config) do
-      StatsD.increment("mail.sent.success", 1, %{mailer: metadata.mailer})
-    end
+      # tracks the number of emails sent successfully/errored
+      defmodule MyHandler do
+        def handle_event([:swoosh, :deliver, :stop], _measurements, metadata, _config) do
+          StatsD.increment("mail.sent.success", 1, %{mailer: metadata.mailer})
+        end
 
-    def handle_event([:swoosh, :deliver, :exception], _measurements, metadata, _config) do
-      StatsD.increment("mail.sent.failure", 1, %{mailer: metadata.mailer})
-    end
+        def handle_event([:swoosh, :deliver, :exception], _measurements, metadata, _config) do
+          StatsD.increment("mail.sent.failure", 1, %{mailer: metadata.mailer})
+        end
 
-    def handle_event([:swoosh, :deliver_many, :stop], _measurements, metadata, _config) do
-      StatsD.increment("mail.sent.success", length(metadata.emails), %{mailer: metadata.mailer})
-    end
+        def handle_event([:swoosh, :deliver_many, :stop], _measurements, metadata, _config) do
+          StatsD.increment("mail.sent.success", length(metadata.emails), %{mailer: metadata.mailer})
+        end
 
-    def handle_event([:swoosh, :deliver_many, :exception], _measurements, metadata, _config) do
-      StatsD.increment("mail.sent.failure", length(metadata.emails), %{mailer: metadata.mailer})
-    end
-  end
+        def handle_event([:swoosh, :deliver_many, :exception], _measurements, metadata, _config) do
+          StatsD.increment("mail.sent.failure", length(metadata.emails), %{mailer: metadata.mailer})
+        end
+      end
 
-  :telemetry.attach_many("my-handler", [
-     [:swoosh, :deliver, :stop],
-     [:swoosh, :deliver, :exception],
-     [:swoosh, :deliver_many, :stop],
-     [:swoosh, :deliver_many, :exception],
-   ], &MyHandler.handle_event/4, nil)
-  ```
+  in `c:Application.start/2` callback:
+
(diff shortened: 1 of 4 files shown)


joshnuss:
Let me know thoughts on naming, maybe event should be named `[:swoosh, :deliver, :success]` and `[:swoosh, :deliver_many, :success]`. 

Alternatively, we should avoid having 2 sets of events, and just pass `%{count: 1}` as metadata when it's only delivering one event

lib/swoosh/mailer.ex
```diff
@@ -106,6 +107,16 @@ defmodule Swoosh.Mailer do
       defp parse_config(config) do
         Mailer.parse_config(@otp_app, __MODULE__, @mailer_config, config)
       end
+
+      defp instrument(response={tag, term}, key, metadata \\ %{}) do
+        status = if(tag == :ok, do: :success, else: :failure)
+        event = [:swoosh, key, status]
+        metadata = Map.merge(metadata, %{mailer: __MODULE__})
+
+        :telemetry.execute(event, %{}, metadata)
```
thiamsantos, line 116: I think would be better if we use [`:telemetry.span/3`](https://hexdocs.pm/telemetry/telemetry.html#span/3). It will emit the events in a more standardized way, following what others libraries do like oban, redix, finch etc. 

And I think we would probably benefit of publishing all the available information in the event, as it allows the listener to use wherever information makes sense to the application and monitoring tool been used.

```elixir
metadata = %{email: email, config: config, mailer: __MODULE__}

:telemetry.span([:swoosh, :deliver], metadata, fn ->
  case Mailer.deliver(email, parse_config(config)) do
    {:ok, result} ->
      {{:ok, response}, Map.put(metadata, :result, result)}

    {:error, reason} ->
      {{:error, reason}, Map.put(metadata, :error, reason)}
  end
end)
```

princemaple:
As @thiamsantos mentioned, `:telemetry.span` can be good.

I think separate events are fine. They may better reveal specific issues like, say, provider's deliver many endpoint is having problem, instead of being mistaken as flaky deliveries.

Yes event names might be better reflect the function names. `deliver` and `deliver_many`

lib/swoosh/mailer.ex
joshnuss: Thanks @thiamsantos, I've updated the PR.

mix.exs
```diff
@@ -60,6 +60,7 @@ defmodule Swoosh.Mixfile do
     [
       {:mime, "~> 1.1 or ~> 2.0"},
       {:jason, "~> 1.0"},
+      {:telemetry, "~> 0.4 or ~> 1.0"},
```
princemaple, line 63:
```suggestion
      {:telemetry, "~> 0.4.2 or ~> 1.0"},
```

span was added in 0.4.2

princemaple:
I pushed a couple commits. I believe with some docs this PR is ready to go in.

joshnuss:
Perfect, nice cleanup of test.

I added the list of telemetry events to the readme. Let me know if you want an example too, or will it clutter the readme?

princemaple:
I think we can add the telemetry example in mailer.ex and link to it in readme.

princemaple:
Thanks @joshnuss !

PR #620, opened 2021-08-21
Build upon async emails section

josevalim:
* Comment that async emails are not necessary
* Comment the downsides of doing async
* Use Task.Supervisor by default


princemaple:
Thanks José!

PR #628, opened 2021-10-11
Adding support for inline attachments replacements in MailboxPreview

theodowling:
I use swoosh for sending emails with inline-attachments. It is difficult to preview the email with images when using MailboxPreview.

I decided to do some string replacement to link to the attachments correctly when viewing in browser. But not sure if it is something we would like to enable by default or just enable with a flag (config, specific-route or query-parameter)?

Please let me know what you think about this please?

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -32,7 +32,7 @@ jobs:
           path: |
             deps
             _build
-          key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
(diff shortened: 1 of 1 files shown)


theodowling:
Hi @princemaple - I've merged your refactor PR. Thanks for having a look and suggesting those edits. 
They look good!

Are you happy to merge or do you want me to review anything else? 

If you also don't mind adding `hacktoberfest-accepted` label to this PR it would be much appreciated 😄

princemaple:
Thanks @theodowling !

One last thing, could you post a screenshot of this feature?

PR #635, opened 2021-11-14
Clarify that you need to add :gen_smtp as a dependency

Hermanverschooten:
I ran afoul of this after deploying my app to production.

The fact that the beam crashed on this with no clear error message took me some time to investigate.


princemaple:
Thank you!

PR #638, opened 2021-11-30
Add provider options doc for Amazon SES adapter

kianmeng:
Beside the provider options, we also fix some markdown to ensure
consistency.

image


princemaple:
Thanks!

PR #639, opened 2021-12-02
Update provider options doc for Mailgun adapter

kianmeng:
We update and standardize the provider options and add a sample code
example on how to use these provider options.

image

### DIFF
--- a/lib/swoosh/adapters/mailgun.ex
+++ b/lib/swoosh/adapters/mailgun.ex
@@ -48,7 +48,7 @@ defmodule Swoosh.Adapters.Mailgun do
     * `:custom_vars` (map) - used to translate to `v:my-var`, now
       `h:X-Mailgun-Variables`, add custome data to email
 
-    * `:recipient_vars` (map | list[map]) - `recipient-variables`, assign
+    * `:recipient_vars` (map) - `recipient-variables`, assign
       custom variable for each email recipient
 
     * `:sending_options` (map) - `o:my-key`, all the sending options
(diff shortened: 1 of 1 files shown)


lib/swoosh/adapters/mailgun.ex
```diff
@@ -21,17 +21,47 @@ defmodule Swoosh.Adapters.Mailgun do
...
 
-  - `:custom_vars` (used to translate to `v:`, now `h:X-Mailgun-Variables`)
-  - `:sending_options` (`o:`)
-  - `:template_name` (template)
-  - `:recipient_vars` (recipient-variables)
-  - `:tags` (`o:tag`, was added in before `:sending_options`)
+    * `:custom_vars` (map) - used to translate to `v:my-var`, now
+      `h:X-Mailgun-Variables`, add custome data to email
+
+    * `:recipient_vars` (map | list[map]) - `recipient-variables`, assign
```
princemaple, line 51: silly question (I haven't used mailgun in a while), what happens when it's a list? how does it map recipients to variables?

kianmeng: I was having the same thoughts as well, assumption of the type for `recipient_vars` made based on these codes:
```
191   defp encode_variable(var) when is_map(var) or is_list(var),
192     do: Swoosh.json_library().encode!(var)
193 
194   defp encode_variable(var), do: var
```

See https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/mailgun.ex#L191

princemaple: https://github.com/swoosh/swoosh/pull/418 I guess list was there to be flexible, but recipient vars should probably be a map.

princemaple:
```suggestion
    * `:recipient_vars` (map) - `recipient-variables`, assign
```

princemaple:
🙏 Thank you again!

princemaple:
FYI I publish the docs directly without making a new release.

PR #647, opened 2021-12-11
Add example to Dyn adapter

kianmeng:


princemaple:
I really wonder how you came up with all the usernames.

kianmeng:
> I really wonder how you came up with all the usernames.

Wikipedia. :laughing:

PR #665, opened 2022-01-15
allow custom CIDs for inline attachments

### LINKED ISSUE #658: Allow using custom CIDs for inline attachments

The docs state that:
> Inline attachments by default use their filename (or basename of the path if filename is not specified) as cid, in relevant adapters.

There appears to be no way to override this. However, you might want to attach multiple files with the same name, or you might not be in control of the email's HTML where CIDs are already defined.

**Describe the solution you'd like**
You could do something like `Swoosh.Attachment.new({:data, attachment_data}, inline: true, cid: "custom-cid")` when building attachments.

**Describe alternatives you've considered**
You can work around the problem in a few ways (like just not attaching multiple files with the same name), but this would be a cleaner solution -- at least for my use case.

**Additional context**
My specific use case is an email forwarding service where CIDs are already set in the email HTML.

taobojlen:
Closes #658 

Added some tests for this (for OhMySMTP and Postmark, the only adapters that handle inline images). Also tested manually with OhMySMTP and it works.

### DIFF
--- a/lib/swoosh/adapters/oh_my_smtp.ex
+++ b/lib/swoosh/adapters/oh_my_smtp.ex
@@ -102,7 +102,7 @@ defmodule Swoosh.Adapters.OhMySmtp do
         }
 
         if attachment.type == :inline do
-          Map.put(attachment_object, :cid, name)
+          Map.put(attachment_object, :cid, attachment.cid)
         else
           attachment_object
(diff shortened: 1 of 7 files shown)


princemaple:
Hey, another great PR. Thanks a lot.

PR #667, opened 2022-01-25
Change assert_email_sent to pass if assigns differ

maxsalven:
`assert_email_sent(email)` currently fails if the `email.assigns` structs differ.

I'm coming from Bamboo, where this is not the case:
https://github.com/thoughtbot/bamboo/issues/154

As an example, I might have

users.exs
```elixir
def grant_admin_permission(user) do
  user = Repo.preload(:permissions)

  # ...grant some permissions based on existing user.permissions

  MyApp.Mailer.admin_permission_granted(user)
  |> MyApp.Mailer.deliver()
end
```

test.exs
```elixir
test "can grant admin permissions and an email gets sent" do
  user = Users.create(%{name: "Sue", email: "sue@example.com"})
  # Note that we don't have the association loaded here in the test
  refute Ecto.assoc_loaded?(user.permissions)

  Users.grant_admin_permission(user)
  assert_email_sent(MyApp.Mailer.admin_permission_granted(user))
end
```

This test will fail, as the email that got delivered had a different `assigns.user.permissions value` than the one in the test.

This PR ignores any assigns when asserting an email was sent, as they're not part of the actual email.

### DIFF
--- a/lib/swoosh/adapters/test.ex
+++ b/lib/swoosh/adapters/test.ex
@@ -21,6 +21,8 @@ defmodule Swoosh.Adapters.Test do
 
   @impl true
   def deliver(email, _config) do
+    email = clean_assigns(email)
+
     for pid <- pids() do
       send(pid, {:email, email})
     end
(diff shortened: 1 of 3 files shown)


princemaple:
Hi Max,

We do intend to test that the sent email is identical to what is being passed in this particular case.

I'm going to close this PR because:
1. It's working as intended.
2. Changing this would be a surprise to existing Swoosh users.
3. You should not write tests this way. You are essentially testing `add(1, 2) == add(1, 2)` but with some side effects, even if it passed it's not testing anything.
4. What you should do, is test the properties that you care about is correctly set/transformed in the email. You can do that by using other forms of `assert_email_sent` like so:
```elixir
assert_email_sent(subject: "Hello, Admin")
```
You can also pass a function and do customized assertion in the function 
https://hexdocs.pm/swoosh/Swoosh.TestAssertions.html#assert_email_sent/1

If you simply wanted to test that an email was sent, you can use `assert_email_sent/0`

PR #668, opened 2022-01-28
Add fields to Postmark `deliver_many` response

zporter:
While working on a feature to send a number of emails using `deliver_many`, I was referencing [Postmark's docs on sending batch emails](https://postmarkapp.com/developer/api/email-api#send-batch-emails). I had assumed that the value returned by the Postmark adapter's `deliver_many` implementation (and also `deliver`) would return the API response using something like `Jason.decode!`, so my original implementation was expecting the String keys matching the casing found in the Postmark API response. I was surprised to find the response was a map with atomized keys and only a subset of the API response. Digging into the Postmark adapter's source, I found the bit of code that was performing the JSON decoding and mapping a subset of values. I need to extend that return value with the "To" field from the API response so I can perform some internal mapping of the recipient and store a receipt of the email.

This PR extends the Postmark adapter's `deliver_many` response to include the "To" and "SubmittedAt" fields from the API. I need the "To" field for my project, but I threw in the "SubmittedAt" field because I figured some additional data couldn't hurt. I'm looking forward to your feedback. Thanks! 💜

### DIFF
--- a/lib/swoosh/adapters/postmark.ex
+++ b/lib/swoosh/adapters/postmark.ex
@@ -112,7 +112,9 @@ defmodule Swoosh.Adapters.Postmark do
             %{
               id: email_result["MessageID"],
               error_code: email_result["ErrorCode"],
-              message: email_result["Message"]
+              message: email_result["Message"],
+              to: email_result["To"],
+              submitted_at: email_result["SubmittedAt"]
             }
           end)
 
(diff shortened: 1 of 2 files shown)


princemaple:
Looks good to me 😉

princemaple:
Will make a release in a few hours.

PR #672, opened 2022-02-08
Migrate OhMySmtp to Mailpace

princemaple:


lib/swoosh/adapters/mail_place.ex
```diff
@@ -0,0 +1,116 @@
+defmodule Swoosh.Adapters.MailPace do
```
taobojlen, line 1: This file is called `mail_place.ex` but should be `mail_pace.ex`.

princemaple: Yup, my bad

test/swoosh/adapters/oh_my_smtp_test.exs
```diff
@@ -1,8 +1,8 @@
-defmodule Swoosh.Adapters.OhMySmtpTest do
+defmodule Swoosh.Adapters.MailPaceTest do
```
kurtome, line 1: Looks like the file name should match the module name

princemaple: haha yep.

princemaple:
Thank you both for the review. @taobojlen @kurtome 

If not more changes are required I'll merge it in a couple hours and release a new version.

PR #674, opened 2022-02-18
Add message_stream documentation to Postmark adapter

ntodd:
`put_provider_option/3` accepts `:message_stream` as an option, but this was not documented.


princemaple:
Thanks 🙏

PR #681, opened 2022-04-05
Rename Mime-Version header to MIME-Version

tcitworld:
Even though [RFC5234](https://www.rfc-editor.org/rfc/rfc5234#section-2.3) says email headers are case-insensitive, prefer to use the value given by [RFC1341](https://datatracker.ietf.org/doc/html/rfc1341#page-5).

For instance, [rspamd doesn't like `Mime-Version`](https://github.com/rspamd/rspamd/blob/f24d59a607066d339597a1a537f3171c9350a1d6/rules/headers_checks.lua#L589-L596).


princemaple:
Thank you. Makes sense.

princemaple:
Released in 1.6.4

Issue #683, opened 2022-04-21
Support for ExAws credentials in SES adapter?


ascandella:
This is not so much a specific feature request, but wanted to leave my thoughts here after spending afternoon getting Swoosh set up with SES for my AWS application.

The current method for configuring SES credentials (e.g. #361) is to pass them in as elixir application configuration (`:access_key` / `:secret`) on the adapter, which would generally happen in `runtime.exs` in a Phoenix application.

This works fine if you use static credentials, but this is not the optimal way to configure AWS permissions for services running inside AWS. Instead of generating (and securely storing / fetching, and maybe possibly even rotating) a single pair of keys that are valid until revoked, you can use the "instance role" (IAM role) of a running service (lambda, EC2 container, ECS service, etc.). Then, when that application needs to talk to another AWS service, it contacts a metadata service running on a special `169.254.169.254` server. This is generally abstracted away by every AWS library (including `ex_aws`).

I understand the reason for not taking on an `ex_aws` dependency for this library, but I imagine at least one other person may run into this limitation (where they don't want to provision a static keypair, instead delegating to the usual IAM strategy).

For my application, I've created a shim adapter that wraps the `AmazonSES` adapter provided by this library:

```elixir
defmodule ExAwsSESAdapter do                                                                                  
  @moduledoc """                                                 
  In order to support instance roles (rather than                                                       
  static keys passed through via environment variables), call `ExAws.Config.new`                                                    
  to respect the settings we've already set up for the app and pass those                                                           
  temporary credentials along to the `Swoosh.Adapters.AmazonSES`.                                                                   
  """                                                                                                                               
  use Swoosh.Adapter,                                                                                                               
    required_deps: [gen_stmp: :mimemail],                                                                                           
    required_config: [:region]                                                                                                      
                                                                                                                                    
  alias Swoosh.Email                                                                                                                
  alias Swoosh.Adapters.AmazonSES                                                                                                   
                                                                                                                                    
  @impl true                                                                                                                        
  def deliver(%Email{} = email, config \\ []) do                                                                                    
    credentials = ExAws.Config.new(:ses)                                                                                            
                                                                                                                                    
    config = add_credentials(config, credentials)                                                                                   
                                                                                                                                    
    email                                                                                                                           
    |> add_security_token(credentials)                                                                                              
    |> AmazonSES.deliver(config)                                                                                                    
  end                                                                                                                               
                                                                                                                                    
  defp add_credentials(config, credentials) do                                                                                      
[...]

PR #684, opened 2022-04-26
Add ExAwsAmazonSES adapter

ascandella:
As discussed in https://github.com/swoosh/swoosh/issues/683

This allows using credentials that are provided by [ExAws](https://github.com/ex-aws/ex_aws), which is especially useful if you're using instance roles where no static access key / secret key pair exists for your service.

It was discussed whether this was worth folding into the existing `AmazonSES` adapter (and auto-detecting if `ExAws` was available), however it was suggested to create a new adapter.


princemaple:
Thank you @ascandella for the excellent PR!

PR #698, opened 2022-06-13
Update Telemetry example to mention errors on `:stop`

lucasmazza:
Failures handled by adapters won't trigger an `:exception` event, but it will include an `:error` key on the `:stop` metadata. 

We noticed that we our log handler wasn't logging errors after a [Mailgun incident](https://status.mailgun.com/incidents/gcm3zw1lvfl4) and the adapter was returning `{:error, term}` instead of raising


princemaple:
Thanks 👍

PR #702, opened 2022-07-14
Support assertions for headers

MatheusBuss:
Added headers clause for `assert_equal` and some tests to check this functionality.

### DIFF
--- a/test/swoosh/test_assertions_test.exs
+++ b/test/swoosh/test_assertions_test.exs
@@ -300,12 +300,12 @@ defmodule Swoosh.TestAssertionsTest do
   end
 
   test "refute email sent with specific header" do
-    refute_email_sent(headers: unquote(%{"Revengers" => "Gather"}))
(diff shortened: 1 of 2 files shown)


test/swoosh/test_assertions_test.exs
```diff
@@ -292,6 +299,16 @@ defmodule Swoosh.TestAssertionsTest do
...
+  test "refute email sent with specific header" do
+    refute_email_sent(headers: unquote(%{"Revengers" => "Gather"}))
+  end
+
+  test "refute email sent with expected header" do
+    assert_raise ExUnit.AssertionError, fn ->
+      refute_email_sent(headers: unquote(%{"Avengers" => "Assemble"}))
+    end
+  end
+
```
MatheusBuss, line 311: These tests may show that it still needs some work, since I couldn't get the tests working without the `unquote`

princemaple: Thanks. I pushed some refactoring on refute. Works now.

PR #709, opened 2022-08-09
Set attachment's ContentId in Mailjet

marcinkoziej:
I can see that cid is not honoured in other adapters as well, but I just checked the Mailjet docs for how to pass that param in there.


princemaple:
Yeah Cid was added to attachment retrospectively. The specific support is then added to the adaptors on demand.

Have you tested attachments that are not inlined?

princemaple:
Tested. Seems to work just fine.

princemaple:
Released in 1.7.4

marcinkoziej:
Thanks! :pray: sorry I did not test, was loaded with work

princemaple:
No worries. Thanks for the PR 🙂

PR #711, opened 2022-08-18
doc: correct tags example for Adapters.Sendinblue

03juan:
The example uses the wrong key and passes a map.


princemaple:
Thank you!

PR #717, opened 2022-09-08
Fix crashes caused by the memory GenServer restarts

KiKoS0:
Connecting/disconnecting nodes to a cluster will cause multiple restarts of the `Memory` adapter GenServer making [exceeding the `max_restarts` of the top-level supervisor](https://hexdocs.pm/elixir/1.13/Supervisor.html#init/2) easy thus crashing the application.

This adds a manager GenServer that will monitor the global adapter instead of linking to it to avoid crashing it when a node disconnects and trigger the above condition.

The bug is triggered by having two nodes in a cluster and running this code:
```elixir
def trigger do
    Node.connect(:"node1@127.0.0.1")
    Process.sleep(400)
    Node.disconnect(:"node1@127.0.0.1")
    Process.sleep(400)
    Node.connect(:"node1@127.0.0.1")
    Process.sleep(400)
    Node.disconnect(:"node1@127.0.0.1")
end
```

Not sure how I can add a test case for this so I'll look into it further first.

### DIFF
--- a/lib/swoosh/adapters/local/storage/manager.ex
+++ b/lib/swoosh/adapters/local/storage/manager.ex
@@ -1,0 +1,39 @@
+defmodule Swoosh.Adapters.Local.Storage.Manager do
+  @moduledoc ~S"""
+  Manages the creation/monitoring of the global in-memory storage driver,
+  `Swoosh.Adapters.Local.Storage.Memory`
+  """
+
+  use GenServer
+
+  def start_link(args \\ []) do
+    GenServer.start_link(__MODULE__, args, name: __MODULE__)
+  end
+
+  @doc false
+  def init(_args) do
+    {:ok, restart_or_monitor()}
+  end
+
+  @doc false
+  def handle_info({:DOWN, _, :process, _, :normal}, _) do
+    # In case the process is stopped normally with `stop/0`
+    {:stop, :normal, %{}}
+  end
+
+  def handle_info({:DOWN, _, :process, _, _reason}, _) do
+    # Try to either restart the global GenServer or monitor the newly
+    # created one.
+    {:noreply, restart_or_monitor()}
+  end
+
+  defp restart_or_monitor() do
+    case Swoosh.Adapters.Local.Storage.Memory.start() do
+      {:ok, pid} -> pid
+      {:error, {:already_started, pid}} -> pid
+    end
+    |> Process.monitor()
+
+    %{}
+  end
+end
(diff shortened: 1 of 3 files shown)


princemaple:
What do you think about simply not starting it by default in prod? It's never meant to be used in prod. We just didn't expect it to cause any issue so it's left there.

KiKoS0:
I believe that would be ideal, but wouldn't that be a breaking configuration change for dev environments that are currently expecting `local: true` by default? Or do you think it's possible through `Mix.env()`?

princemaple:
Yeah compile with different functions or function body depending on `Mix.env()`.

princemaple:
```elixir
@local_default Mix.env() == :dev

defp runtime_children(children, :local) do
  if Application.get_env(:swoosh, :local, unquote(@local_default)) do
    ...
```

KiKoS0:
The issue with that approach is that dependencies are always compiled in production mode.
https://elixirforum.com/t/mysterious-magical-prod-option-in-mix-exs/38036/2

princemaple:
Huh. TIL.

lib/swoosh/adapters/local/storage/manager.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  @doc false
+  def handle_info({:DOWN, _, :process, _, :normal}, _) do
+    # In case the process is stopped normally with `stop/0`
+    {:stop, :normal, %{}}
+  end
+
+  def handle_info({:DOWN, _, :process, _, _reason}, _) do
+    # Try to either restart the global GenServer or monitor the newly
+    # created one.
```
princemaple, line 26: Is it worth adding a random duration sleep here?

KiKoS0: While this change already reduces the number of restarts, I think it wouldn't hurt to have a delay in case connects/disconnects are happening fast. Do you think a maximum of 2 seconds `Process.sleep(:rand.uniform(1_000) + 1_000)` is fine?

princemaple: Maybe something like 200 + rand1800 to maximize variation. What do you think?

KiKoS0: Sounds good, I'll add it.

princemaple: 🙏

princemaple:
Released 1.8.0

KiKoS0:
Awesome, thank you for the review and the really fast release 🎉

princemaple:
❤️ thank you for the detailed issue and companion PR :)

PR #718, opened 2022-09-12
Fix empty mail "To"

barrosodegas:
## Fixes email recipient validation.
- Adds email recipient validation when not informed.

### DIFF
--- a/lib/swoosh/delivery_error.ex
+++ b/lib/swoosh/delivery_error.ex
@@ -7,6 +7,7 @@ defmodule Swoosh.DeliveryError do
   end
 
   defp format_error(:from_not_set, _), do: "expected \"from\" to be set"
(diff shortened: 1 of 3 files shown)


princemaple:
🤔 

There are 2 problems with this.

1. Emails built with Swoosh.Email functions have `to` as a list. It will never be nil or a single tuple.
2. This, while implemented properly (checking empty list and every item of the list), will be a potential breaking change. Some users may be relying on the behaviour that empty `to` list not sending to anyone.

It's probably best implemented in your own codebase as a custom deliver function that checks the `to` before calling the original mailer deliver function.

princemaple:
Actually our code already checks whether the recipients are nil or empty string. So the only case we are not checking is empty list, which is intentional.

princemaple:
Thanks for the PR. I'm going to close it now.

PR #721, opened 2022-10-04
Swoosh.Adapters.Test.delivery_many/2 returns a list

markthequark:
Update the `Swoosh.Adapter.deliver_many/2` typespec to return a list and update the `deliver_many/2` implementation in `Swoosh.Adapter.Test`.

This better represents how the current adapters, Mailjet and Postmark, implement the `deliver_many/2` callback.


princemaple:
I'm currently travelling. It will take a while before I can get to this. Thanks for the PR.

princemaple:
I'm back from my trip. I'll get to this in a couple days. Thanks for the patience.

princemaple:
I was hesitant to merge this because it's going to be a breaking change. After some thought, I decided to proceed with this as is, because it does indeed make more sense. The previous behavior was pretty useless, and hopefully isn't used by anyone. Thanks a lot!

PR #722, opened 2022-10-07
Postmark: Support per email tracking options

Wijnand:
Postmark allows overriding the "Open" and "Click" tracking on a per email basis. This commit allows (optionally) setting the required fields in the API calls via the provider options.


princemaple:
Thanks 🙏

PR #734, opened 2022-12-16
[ADAPTER] customer.io 

lucacorti:
Hi,

I need to send emails via customer.io transactional email API. This implementation might have some rough edges but is working for me.

### DIFF
--- a/lib/swoosh/adapters/customer_io.ex
+++ b/lib/swoosh/adapters/customer_io.ex
@@ -98,6 +98,8 @@ defmodule Swoosh.Adapters.CustomerIO do
 
   alias Swoosh.Email
 
+  import Swoosh.Email.Render
+
   @base_url "https://api.customer.io/v1"
   @api_endpoint "/send/email"
   @provider_options_body_fields [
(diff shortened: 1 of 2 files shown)


lib/swoosh/adapters/customer_io.ex
```diff
@@ -0,0 +1,214 @@
...
+    Map.put(body, :headers, headers)
+  end
+
+  defp prepare_provider_options_body_fields(body, %Email{provider_options: provider_options}) do
+    Map.merge(body, Map.take(provider_options, @provider_options_body_fields))
+  end
+
+  defp email_item({"", address}), do: address
+  defp email_item({name, address}), do: name <> " <" <> address <> ">"
+  defp email_item(address), do: address
```
princemaple, line 213: https://github.com/swoosh/swoosh/blob/main/lib/swoosh/email/render.ex

Does this work?

princemaple:
Fairly good looking PR, thanks! Just that one question for now.

lib/swoosh/adapters/customer_io.ex
lucacorti: Yes, I just copied the code over from the Sendgrid adapter, fixed.

princemaple:
🎉Thank you!

PR #746, opened 2023-02-26
Allow overriding delivery functions

jakub-gonet:
Sometimes, it's really convenient to do some preprocessing of an email (returning an error when recipient list is empty, inlining CSS, etc) that is adapter independent.

I'm open to suggestions how to improve telemetry as only base implementation is covered by it.

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -176,6 +176,8 @@ defmodule Swoosh.Mailer do
           end
         end)
       end
+
+      defoverridable deliver: 2, deliver!: 2, deliver_many: 2
     end
   end
 
--- a/test/swoosh/mailer_test.exs
+++ b/test/swoosh/mailer_test.exs
@@ -131,6 +131,20 @@ defmodule Swoosh.MailerTest do
     end
   end
 
+  test "allows overriding default delivery functions", %{valid_email: email} do
+    defmodule OveriddenMailer do
+      use Swoosh.Mailer, otp_app: :swoosh, adapter: FakeAdapter
+
+      def deliver(%Swoosh.Email{to: []}, _config), do: {:error, :no_recipient}
+      def deliver(email, config), do: super(email, config)
+    end
+
+    email_without_recipient = %{email | to: []}
+
+    assert {:ok, _} = OveriddenMailer.deliver(email)
+    assert {:error, :no_recipient} = OveriddenMailer.deliver(email_without_recipient)
+  end
+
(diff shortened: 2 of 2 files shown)


princemaple:
Thanks for the PR!

I'm in general not in favor of unnecessary overrides and hooks. I think preprocess can be done with a wrapper function, e.g. `safe_deliver(...)` that wraps existing function. Inlining CSS sounds more like a rendering problem which should be solved before it is put into the email, at the template (engine) level.

I'm not strongly against making them overridable though. 🤔

princemaple:
Actually yeah I don't want to make them overridable. This would cause more magic and make things confusing and hard to search for. I think I much prefer keeping it as is and have users write wrapper functions - so that later on when they have doubts about what their custom `deliver_*` or `*_deliver` does, they get the correct search result.

💜 for the PR though!

jakub-gonet:
Sure, thanks for feedback!

jakub-gonet:
One another though I had this morning about this: I felt like doing it, because Swoosh injects functions in module scoped to my project.

If I use a library, I expect all calls to it directly to have same behavior but any call to `MyProject.Library` is in uncharted territory and if I had a reason to use this instead of library directly then I probably added some additional logic.

That being said, I'm fine with this PR being rejected.

princemaple:
I looked at how `Ecto` does `Repo`. Seeing that `all` and `one` and the rest aren't overridable, I decided that `Mailer` shouldn't make `deliver` overridable either. One could argue they'd want to override `all` to make a cached version, but again, one could always just make a `cached_all`.

My own opinion is really just code discoverability. Code should be easily searchable and app code should not be confused with library code.

adamu:
I think it might be better if instead of injecting the `deliver` functions from `use Swoosh.Mailer`, they were actually defined on the Swoosh.Mailer module (or a behaviour) that library users could choose the call as they like. For example, in the current version, the telemetry logic is [only applied by default when using the injected functions](https://github.com/swoosh/swoosh/blob/4a59f125c014edaaf9f23de43f76db1e8cb3b691/lib/swoosh/mailer.ex#L131-L133).

princemaple:
The major reason for code injection is to read the specific configuration for the mailer, through the module name.

However I agree that the injected code can be further reduced, as noted in https://github.com/swoosh/swoosh/issues/556

Maybe when we get to a point where the injected code is reduced to the bare minimum, and `use Swoosh.Mailer` becomes a pure convenience like `use GenServer`, i.e. contains no code with logic, we can make the deliver functions overridable.

PR #747, opened 2023-02-26
Add experimental new test assertion module

jakub-gonet:
❗ This isn't backwards compatible.

I had a few complaints about current implementation of test helpers - `refute` working only on literal params, assertions consuming emails from queue, kinda bad error reporting so I decided to take a spin on it.

It's a bit worse memory-wise because we keep entire email list but it shouldn't affect anything, I don't expect people to create MB worth of emails in tests.

I also thought about removing delegating assertions (`assert_email_not_sent` and family) since this is breaking change anyway.

Similar tests were grouped, I think it improves readability a bit.

I would love to hear your thoughts about this one.

### DIFF
--- a/lib/swoosh/x_test_assertions.ex
+++ b/lib/swoosh/x_test_assertions.ex
@@ -265,24 +265,6 @@ defmodule Swoosh.X.TestAssertions do
     """
   end
 
-  @doc ~S"""
-  Asserts no emails were sent.
-  """
-  @spec assert_no_email_sent() :: boolean | no_return
-  def assert_no_email_sent() do
-    refute_email_sent()
-  end
-
-  @doc ~S"""
-  Asserts `email` was not sent.
-
-  Performs exact matching of the email struct.
-  """
-  @spec assert_email_not_sent(Email.t()) :: boolean | no_return
-  def assert_email_not_sent(%Email{} = email) do
-    refute_email_sent(email)
-  end
-
   @doc """
   Removes and returns from mailbox all sent emails.
   """
--- a/lib/swoosh/x_test_assertions.ex
+++ b/lib/swoosh/x_test_assertions.ex
@@ -1,0 +1,309 @@
+defmodule Swoosh.X.TestAssertions do
+  @moduledoc ~S"""
+  This module contains a set of assertions functions that you can import in your
+  test cases.
+
+  It is meant to be used with the
+  [Swoosh.Adapters.Test](Swoosh.Adapters.Test.html) module.
+
+  **Note**: `Swoosh.TestAssertion` works for unit tests and basic integration tests.
+  Unfortunately it's not going to work for feature/E2E tests.
+  The mechanism of `assert_email_sent` is based on messaging sending between processes,
+  and is expecting the calling process (the one that calls `assert_email_sent`) to be
+  the calling process of `Mailer.deliver`, or be the parent process of the whatever
+  does the `Mailer.deliver` call.
+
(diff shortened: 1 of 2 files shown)


test/swoosh/test_assertions_test.exs
```diff
@@ -123,25 +92,22 @@ defmodule Swoosh.TestAssertionsTest do
...
   test "assert email not sent with unexpected email" do
     unexpected_email = new() |> subject("Testing Avenger")
+
     assert_email_not_sent(unexpected_email)
   end
 
   test "assert email not sent with expected email", %{email: email} do
-    message =
-      "Unexpectedly received message {:email, #{inspect(email)}} (which matched {:email, ^email})"
+    message = ~r/Expected[\s\S]+to not contain[\s\S]+but this email matched/
```
princemaple, line 100: `[\s\S]+` is just `.+`? 😄

princemaple:
This has some good changes. Since it's a breaking change, it's probably never going to make it into 1.x so I think I have a lot of time to review it and test it. Your patience would be most appreciated! 😄

test/swoosh/test_assertions_test.exs
jakub-gonet: `.` doesn't match newlines we have in messages, `[\s\S]` is the next best thing.

jakub-gonet:
It can live here for a while, I don't mind it.

We probably could think of some plan to make it backwards compatible by renaming functions or moving it into other module you can import (and disable previous one by `import Swoosh.TestAssertions, only: []`) but it feels natural to make this kind of breaking change eventually.

test/swoosh/test_assertions_test.exs
princemaple: Thanks for the clarification. Have you tried `~r/Expected.+to not contain/m`? Note the `m` at the end, the `multiline` modifier.

princemaple: Actually my bad. Should be `s`.

princemaple:
Eventually yes, it's easy to throw a `v2` out. However, maybe we don't have to. Was thinking maybe we can have a temporary TestAssertionV2 module for a while, then merge back into `TestAssertion` and deprecate `TestAssertionV2` when swoosh v2 is officially out.

jakub-gonet:
> then merge back into TestAssertion

This is inherently breaking unfortunately. If we add v2 then it'll be most likely undiscoverable since we import original module in macros.

I'll probably find some time in the near future to contribute so if you have vision how you want evolve the project, let me know (you can shoot me an email to `jakub.gonet0@gmail.com` or write it here if it's short). I have some ideas I want to try but I'm more experienced in Elixir than in emails :>

princemaple:
💜 Appreciate the offering.

> This is inherently breaking unfortunately

Yes, that's why I wouldn't change the original `TestAssertion` module in Swoosh v1.

Was thinking about taking this change in, but as a `TestAssertionV2` experimental module in Swoosh v1, which will replace `TestAssertion` when Swoosh v2 is out.

So users who are eager to use this can import from `TestAssertionV2` when Swoosh is still v1. And after Swoosh v2 is out, they can simply drop the `V2` on their test import. And users that are still on original `TestAssertion` will finally have to update.

jakub-gonet:
Ah, now I see. Yeah, that makes sense.

princemaple:
Sorry about not getting to this for so long.

princemaple:
Note: need to bring in #757 and #764

Issue #560, opened 2021-01-09
Swoosh.TestAssertions do not account for deliver_many.


zorn:
Working in my own side project today I was trying to use `assert_email_sent` after a `deliver_many` call. I don't think this is working. Might be doing it wrong. Not sure. My current test is something like this:

```elixir
  describe "send_email_notification_batch/1" do
    setup [:with_various_delivery_status]

    test "success: emails we sent", %{group: group} do
      notes = Notifications.list_email_notifications_ready_for_delivery(prefix: group.subdomain)

      {:ok, _} = Notifications.send_email_notification_batch(prefix: group.subdomain)

      for note <- notes do
        assert_email_sent(Guildflow.NewMessageEmailNotification.new(note))
      end
    end
  end
```

And the failure is a bad match:

<img width="1904" alt="Screen Shot 2021-01-09 at 6 04 11 PM" src="https://user-images.githubusercontent.com/52168/104110424-6efba880-52a5-11eb-9e5a-89eee4937ba6.png">

Poking around the code I'm not sure `Swoosh.TestAssertions` ever expect that kind of message.

princemaple:
Right. Currently `Swoosh.TestAssertions` is not equipped with the necessary tools to deal with `deliver_many`. As a workaround you can just go `assert_receive` until something like `assert_emails_sent` is added. Note, `deliver_many` will not behave like multiple calls to `deliver`, it is meant to be different.

geeksilva97:
Can I work on that? :eyes:

princemaple:
Yeah. https://github.com/swoosh/swoosh/pull/721 take this into consideration, though.

geeksilva97:
Okay.

PR #757, opened 2023-04-07
Add assert_emails_sent

geeksilva97:
This PR adds the `assert_emails_sent`  function (#560 ).

### Proposed changes

- add a function to make it easier to test `deliver_many`

### DIFF
--- a/lib/swoosh/test_assertions.ex
+++ b/lib/swoosh/test_assertions.ex
@@ -23,7 +23,7 @@ defmodule Swoosh.TestAssertions do
   alias Swoosh.Email
   alias Swoosh.Email.Recipient
 
-  @type assert_email :: Email.t() | Keyword.t() | (Email.t() -> boolean())
+  @type email_assertion :: Email.t() | Keyword.t() | (Email.t() -> boolean())
 
   @doc """
   Sets Swoosh test adapter to global mode.
@@ -72,7 +72,7 @@ defmodule Swoosh.TestAssertions do
     assert_received {:email, _}
   end
 
-  @spec assert_email_sent(assert_email()) ::
+  @spec assert_email_sent(email_assertion()) ::
           :ok | tuple | no_return
 
   @doc ~S"""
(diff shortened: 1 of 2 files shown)


lib/swoosh/test_assertions.ex
```diff
@@ -23,6 +23,8 @@ defmodule Swoosh.TestAssertions do
   alias Swoosh.Email
   alias Swoosh.Email.Recipient
 
+  @type assert_email :: Email.t() | Keyword.t() | (Email.t() -> boolean())
```
princemaple, line 26: How about `email_assertion`?

lib/swoosh/test_assertions.ex
```diff
@@ -112,6 +114,52 @@ defmodule Swoosh.TestAssertions do
...
+          :ok | tuple | no_return
+  def assert_emails_sent([%Swoosh.Email{} | _] = emails) do
+    assert_received {:emails, ^emails}
+  end
+
+  def assert_emails_sent([%{} | _] = params_map_list) do
+    assert_received {:emails, emails}
+
+    if length(emails) != length(params_map_list),
+      do: raise("The arguments must have the same length")
```
princemaple, line 154:
```suggestion
    assert length(emails) == length(params_map_list)
```

Is this better or worse? 🤔

princemaple (changes requested):
Looks good in general :) Thanks heaps for the PR

lib/swoosh/test_assertions.ex
geeksilva97: Sounds good

lib/swoosh/test_assertions.ex
geeksilva97: Good question. I'm wondering whether using `assert` could lead to confusion since it's an internal assertion in an assert (in this case `assert_emails_sent`.

princemaple: Thought "The arguments" was kinda vague

princemaple:
💜

geeksilva97:
> Looks good in general :) Thanks heaps for the PR

No prob!

PR #759, opened 2023-04-11
Add postmark inline css option

matehat:
This is useful in some situation where Postmark's default CSS inlining does not yield great results.


princemaple:
Thanks!

PR #764, opened 2023-04-26
Allow Regexp assertions for subjects

aronisstav:
Hi!

I am still a beginner in Elixir/Phoenix/Swoosh, so it took me a while to understand why a regexp-based assertion didn't work with e-mail subject.

I found out, as you can tell, and I think it should be supported, hence this PR!

Thank you for a very beginner-friendly library!

/Stavros

### DIFF
--- a/lib/swoosh/test_assertions.ex
+++ b/lib/swoosh/test_assertions.ex
@@ -159,6 +159,9 @@ defmodule Swoosh.TestAssertions do
     end)
   end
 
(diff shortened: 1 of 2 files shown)


princemaple:
Could you please add a test for it? Thanks :)

princemaple:
I'll make a release maybe tomorrow. Remind me if I forgot to do so :)

Issue #768, opened 2023-05-18
inline attachment showing up as attachment


Hermanverschooten:
### Describe the bug

I am sending a QR-code as an inline attachment, as shown in the documentation.
Upon receipt however the QR-code is show in the HTML - what I want
But it is also shown as a normal attachment, so on iPhone, iPad, Mac Mail etc, it appears again at the bottom of the mail.
I am sending using the SMTP-adapter.

```elixir
new()
|> to(...)
...
|> attachment(
  Swoosh.Attachment.new(
    {:data, gen_qr(data)},
    filename: "qr.png",
    type: :inline,
    content_type: "image/png"
  )
)
```

### Steps to Reproduce the Bug or Issue

Send an e-mail with an inline attachment.

### Expected behavior

The headers should be set so the mail client recognizes it as an inline attachment.

### Platform

- OTP: 25.3.2
- Elixir: 1.14.4

Hermanverschooten:
It seems that in `lib/swoosh/adapters/smtp/helpers.ex` in `prepare_parts/3` a distinctinction should be made if there are inline attachments, they need to be grouped together with the html body parts into a `multipart/relative` section.

princemaple:
Interesting, thank you for your effort in finding out why and a possible solution.

Are you working on a fix? If not I will take a look maybe next week.

Hermanverschooten:
I took a stab at it in #769

Hermanverschooten:
Hmm, I get a failing test on Erlang 26, newly added?

princemaple:
My bad. Feel free to ignore the OTP26 failures for now.

Hermanverschooten:
I already put in a hacky solution for 26...

princemaple:
No worries. The PR looks good. I'll work on the fix for tests.

Hermanverschooten:
You want me to remove the test commit?

Hermanverschooten:
too late, thanks for merging

princemaple:
:) I'm working on the test cleanup now.

princemaple:
Released 1.10.3 for this. Thank you heaps for the detailed report, investigation and taking the initiative to fix it.

PR #769, opened 2023-05-19
When adding an inline-attachment an extra section is needed.

Hermanverschooten:
Mail clients expect an inline attachment to be grouped with
the HTML in a "multipart/related" section, if not the
attachment will show up as both "inline" and as a regular
attachment.


princemaple:
close #768

PR #773, opened 2023-06-09
Do not include Bcc header in delivered email

adamu:
Proposal following [this discussion](https://github.com/swoosh/swoosh/discussions/772).

According to [RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.3) method 1:

> when a message containing a "Bcc:" field is prepared to be sent, the "Bcc:" line is removed even though all of the recipients (including those specified in the "Bcc:" field) are sent a copy of the message

This PR removes the BCC header from the message that is delivered.

Note that the BCC addresses are still included in the `recipients`, so they will still receive the email. It is only the `body` that does not contain the bcc addresses:

https://github.com/swoosh/swoosh/blob/3cab4d812dfef1c842c127e4bb36b475716febe4/lib/swoosh/adapters/smtp.ex#L55-L59

### DIFF
--- a/lib/swoosh/adapters/smtp/helpers.ex
+++ b/lib/swoosh/adapters/smtp/helpers.ex
@@ -51,8 +51,8 @@ defmodule Swoosh.Adapters.SMTP.Helpers do
     |> prepare_mime_version()
     |> prepare_reply_to(email)
     |> prepare_subject(email)
-    |> prepare_bcc(email)
     |> prepare_cc(email)
+    # bcc is deliberately omitted: https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.3
     |> prepare_to(email)
     |> prepare_from(email)
   end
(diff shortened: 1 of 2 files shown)


adamu:
Ah sorry I didn't run all of the tests. Will fix.

princemaple:
💜

adamu:
Digging some more, it looks like this was deliberately added in #105. There might be multiple use-cases? At least the comment I left explains why it's not there now.

thiamsantos:
@princemaple Hello, this change broke our implementation after the update, we had to rollback to `v1.11.0`. 

We use the adapter `Swoosh.Adapters.AmazonSES`, one of our emails do not have any `to` but only the `bcc` fields. I imagine that as the adapter uses the smtp helper to build a email and send via http to aws, SES stopped receiving the `bcc` and cannot build the receipients.

We started receiving from aws.
````
 {:error, %{code: "InvalidParameterValue", message: "Missing required header 'To'."}}
```

adamu:
It looks like this could be the case for the Sendmail adapter too. I guess they should not be using the SMTP Adapter's helper code.

princemaple:
@thiamsantos looks like we have to distinguish between adapters that use SMTP format and real SMTP on handling of BCC. Thanks for reporting. Will fix soon.

PR #776, opened 2023-06-16
Remove unnecessary sentence from README

adamu:
In my project, I want to customise the behaviour of `Swoosh.Mailer`. For example, doing things like:

* Adding custom error handling/reporting
* Chunking delivery of a mail with many recipients into separate mails
* etc...

However, the library makes this difficult, because multiple `deliver` functions are [automatically injected via `use`](https://github.com/swoosh/swoosh/blob/4a59f125c014edaaf9f23de43f76db1e8cb3b691/lib/swoosh/mailer.ex#L125-L155) and are not [overridable](https://hexdocs.pm/elixir/Kernel.html#defoverridable/1). Without being able to override these, the best that can be done is to wrap these functions, however they cannot be replaced so it may be that the default versions can be called accidentally, which I want to avoid (I'd rather replace them with my own versions).

This PR makes the functions overridable, so they can be extended.

There is probably a better architectural solution to this. For example, making `Swoosh.Mailer` a behaviour. However, this was the simplest approach I could think of.

I also removed a self-congratulatory sentence from the README 🙄

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,8 @@
 
 Compose, deliver and test your emails easily in Elixir.
 
-We have applied the lessons learned from projects like Plug, Ecto and Phoenix
-in designing clean and composable APIs, with clear separation of concerns
-between modules. Swoosh comes with many adapters, including SendGrid, Mandrill,
-Mailgun, Postmark and SMTP. See the full list of [adapters below](#adapters).
+Swoosh comes with many adapters, including SendGrid, Mandrill, Mailgun, Postmark and SMTP.
+See the full list of [adapters below](#adapters).
 
 The complete documentation for Swoosh is [available online at HexDocs](https://hexdocs.pm/swoosh).
 


adamu:
The issues I'm having might be specific to SMTP. If you are not happy with this change, I could probably solve the issue by wrapping the SMTP adapter instead (and maybe contributing the changes if they are generic enough).

princemaple:
Hi @adamu 

https://github.com/swoosh/swoosh/pull/746
Yeah I think it's better to use a wrapper mailer / customer adapter.

Also, it's entirely possible to build a mailer without using `use Swoosh.Mailer`. All the building blocks are in public functions in `Swoosh.Mailer`. There is very little code being injected.

Cheers.

BTW happy to accept the readme change though. It's there way before I got myself involved in this project :)

adamu:
@princemaple understood! It feels a bit frivolous to make a PR just to remove a PR from the readme, but have made the change :)

princemaple:
💜

princemaple:
Doc change is out :)

PR #779, opened 2023-06-19
Fix BCC for adapters that depend on SMTP helper

princemaple:
@adamu @thiamsantos 

https://elixirforum.com/t/elixir-v1-15-0-released/56584/4

### DIFF
--- a/lib/swoosh/adapters/smtp/helpers.ex
+++ b/lib/swoosh/adapters/smtp/helpers.ex
@@ -12,7 +12,7 @@ defmodule Swoosh.Adapters.SMTP.Helpers do
 
   @doc false
   def body(email, config) do
-    {message_config, config} = Keyword.split(config, [:transfer_encoding])
+    {message_config, config} = Keyword.split(config, [:transfer_encoding, :keep_bcc])
     {type, subtype, headers, parts} = prepare_message(email, message_config)
     {encoding_config, _config} = Keyword.split(config, [:dkim])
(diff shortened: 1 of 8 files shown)


lib/swoosh/adapters/smtp/helpers.ex
```diff
@@ -41,18 +41,26 @@ defmodule Swoosh.Adapters.SMTP.Helpers do
...
     |> prepare_additional_headers(email)
     |> prepare_mime_version()
     |> prepare_reply_to(email)
     |> prepare_subject(email)
     |> prepare_cc(email)
-    # bcc is deliberately omitted: https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.3
+    |> then(fn headers ->
+      # bcc is deliberately omitted unless explicitly asked to keep, for service providers
+      # https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.3
+      if config[:keep_bcc] do
```
adamu, line 58: I think it might be worth pulling this `body` function out of the SMTP Adapter, maybe into the Adapter module, or creating Adapter.Helpers, WDYT?

I can possibly look into it if you'd like to apply this fix for now, but it probably won't be until next week or later.

princemaple: It's in the helper module, not in the adapter module. Not to say that it's good, I dislike the helper module, but it's not as bad as being in the adapter directly.

PR #781, opened 2023-06-20
Fix sections on CHANGELOG

nelsonmestevao:
The 1.7.5 version had the subsection levels wrong (h2 instead of h3).

See:
image


princemaple:
💜

PR #782, opened 2023-06-25
rename sendinblue -> brevo

garlandcrow:
When I was comparing email api providers, I noticed sendinblue has gone through a company rename to Brevo, so updated the adapters to reflect that.


princemaple:
Hi, thanks for the pr. It contains too much irrelevant change and doesn't handle backwards compatibility. I'll leave it here for a couple days and make necessary changes when I get the time. Cheers

PR #787, opened 2023-07-06
Update Mailgun docs for sandbox mode

stjhimy:
The work is based on this topic: https://elixirforum.com/t/sending-email-404-not-found-nothing-is-happening-no-debug-messages/55230

When using Mailgun sandbox mode you have to set the domain based on the sandbox address used on the API url. Let me know if I should rephrase or move it to somewhere else.

Thanks.


princemaple:
Thanks!

PR #793, opened 2023-07-13
Deps checked during compilation

princemaple:
cc @josevalim


princemaple:
I discovered that with this code, if I

- have all the deps ready
- mix compile
- remove a dep and do mix deps.clean --unused
- mix again -> no error

because the user code isn't recompiled, the before compile hook isn't run.

This probably doesn't affect CI deployment, as you normally recompile your app from clean state, given that you don't cache too heavily.

However, this can be a source of error. Any suggestion?

josevalim:
Oh, the adapter is specified via a config file, that's why compilation time never sees it. I have missed that context so I think this is good as is, no further change necessary. Thank you for looking into this very promptly!

PR #800, opened 2023-08-09
Deprecate system env tuples

josevalim:


lib/swoosh/mailer.ex
```diff
@@ -204,27 +206,33 @@ defmodule Swoosh.Mailer do
...
 
-  @doc """
-  Parse configs in the following order, later ones taking priority:
-
-  1. mix configs
-  2. compiled configs in Mailer module
-  3. dynamic configs passed into the function
-  4. system envs
-  """
+  @doc false
```
josevalim, line 209: I imagine these functions are actually private? :)

princemaple: I forgot why I made them public. Probably just so that I could document them.

PR #801, opened 2023-08-09
Use concatenation to build sup children

josevalim:
Minor change but I hope it simplifies the code. :)

### DIFF
--- a/lib/swoosh/application.ex
+++ b/lib/swoosh/application.ex
@@ -46,13 +46,8 @@ defmodule Swoosh.Application do
         Could not start preview server.
 
         Please add :plug_cowboy to your dependencies:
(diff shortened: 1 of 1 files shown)


lib/swoosh/application.ex
```diff
@@ -40,30 +34,24 @@ defmodule Swoosh.Application do
...
         Could not start preview server.
 
         Please add :plug_cowboy to your dependencies:
-          {:plug_cowboy, ">= 1.0.0"}
-
-        And then recompile swoosh:
 
-          mix deps.get
-          mix deps.clean --build swoosh
-          mix deps.compile swoosh
```
josevalim, line 62: When a dependency is added, Elixir automatically recompiles everything that optionally depends on it. :)

PR #802, opened 2023-08-09
Compute docs lazily

josevalim:
The documentation code traverses the adapters folder,
so we want to perform it lazily and not every time
mix.exs is loaded.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -24,7 +24,7 @@ defmodule Swoosh.Mixfile do
       name: "Swoosh",
       source_url: @source_url,
       homepage_url: @source_url,
-      docs: docs(),
+      docs: &docs/0,
       preferred_cli_env: [
         docs: :docs,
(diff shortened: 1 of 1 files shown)


princemaple:
TIL. Thanks 🙏

krns:
Hi, I just saw this change and wondered why the docs are now computed lazily. Where can I read more about this topic?

princemaple:
https://hexdocs.pm/ex_doc/Mix.Tasks.Docs.html

> ExDoc also allows configuration specific to the documentation to be set. The following options should be put under the :docs key in your project's main configuration. The :docs options should be a keyword list or **a function returning a keyword list** that will be lazily executed.

Swoosh `mix.exs` traverses the adapters dir to find all the adapters. When it was eager loaded it happens every time `mix.exs` is loaded. Jose changed it to lazy load so that it only happens when compiling docs.

Issue #744, opened 2023-01-11
Microsoft Graph support


hermanius:
Are there (short-term) plans to support a Microsoft Graph adapter?
https://learn.microsoft.com/en-us/graph/api/user-sendmail?view=graph-rest-1.0&tabs=http

princemaple:
No plans, happy to accept PR 😊

geeksilva97:
Can I take a look?

princemaple:
😃 sure

justindotpub:
I have a working implementation for this and was pondering if a PR would be accepted.

Would you accept a PR that only supports auth using oauth 2.0 tokens, and that shields swoosh from that logic by having the config be `access_token_fn`, an anonymous function that returns the access token to be used in the authorization header?  This allows the user of this adapter to handle oauth 2.0 access token requests in whatever manner they see fit, including dealing with caching tokens and refreshing before their expiration.

justindotpub:
🤔 I'm pondering how to incorporate that into the code in a generic way.  As long as authorization details are always ultimately incorporated into the `authorization` header I could see this working simply.  I may have to do some research.

princemaple:
Thanks for being rigorous Justin. Feel free to send a PR anyway, in case people want to help you with that. We can iterate on things.

justindotpub:
If you don't mind humoring me, I'd like to think aloud to see if I'm on the right track.  :)

If I end up being correct that auth details are always ultimately provided in the `authorization` header, we could define a new behaviour (e.g. `Swoosh.AuthorizationHeaderProvider`) and then `Swoosh.Adapters.MsGraph` could accept something like `authorization_header_provider` config.  Then in `Swoosh.Adapters.MsGraph` I would use this module to call its `authorization_header` callback.  The advantage of this approach over an anonymous function would primarily be that we call a function on a module with a defined contract, and make it easier for others to reuse various implementations of this behaviour instead of always defining their own anonymous function.

Any corrections or additional thoughts to that?

justindotpub:
Also note that I'm intentionally trying to keep all auth config params out of Swoosh, so there would not be any additional config options for, say, cases where someone is authenticating with username and password.

justindotpub:
Good naming feedback, thank you.

I learned from https://learn.microsoft.com/en-us/graph/auth/auth-concepts that Graph auth always uses oauth 2.0 access tokens.  

[...]

PR #815, opened 2023-09-21
Implement Swoosh.Adapters.MsGraph

justindotpub:
As discussed in https://github.com/swoosh/swoosh/issues/744.


test/support/conn_parser.ex
```diff
@@ -8,7 +8,7 @@ defmodule Swoosh.ConnParser do
         Plug.Parsers.JSON
       ])
       |> Keyword.put_new(:json_decoder, Swoosh.json_library())
-      |> Keyword.put_new(:pass, ["message/rfc822"])
+      |> Keyword.put_new(:pass, ["message/rfc822", "text/plain"])
```
justindotpub, line 11: Without this, `conn = parse(conn)` in tests would blow up.

princemaple:
One last thing, is plug needed for this adapter? I don't see it being used other than in the tests.

justindotpub:
Good catch. I didn’t use it explicitly but assumed it was some requirement under the covers, perhaps due to the use of the api client or something. Let me know I need to make that change. I’ll assume you are otherwise, so I don’t mess up what you’re doing.

justindotpub:
I confirmed, the mailgun adapter that I used as my starting point explicitly used `Plug.Conn.Query.encode`.  I don't need Plug.

princemaple:
Thank you @justindotpub ! This is a great PR.

justindotpub:
Thank you for the support @princemaple! ❤️

PR #825, opened 2023-10-18
Add Scaleway adapter

andreh11:
This PR adds an adapter for Scaleway (transactional emails only).

Ref.: https://www.scaleway.com/en/developers/api/transactional-email/


princemaple:
Thank you @andreh11 , this looks good. I'll have a more thorough review in a few days and merge it then :) Didn't know scaleway has an Email service now. Its docs says `v1alpha1`, do you know if it's stable already or when it will be?

andreh11:
Well, I've been using it in production for only 2 weeks. No problem so far, but this is not a proof of stability. I've just asked Scaleway's support, we'll see. Maybe we can wait for their answer before merging.

Also, the API documentation says nothing about 4xx and 5xx errors, so the related tests are just copy/pasted from `brevo_test.exs`. I'd be happy to improve this part, but how?

andreh11:
I've asked Scaleway about `v1alpha1`. They've confirmed that the API is stable. They won't change the endpoint's URI.

PR #827, opened 2023-10-24
Implement Mailtrap adapter

kalys:
This PR adds an adapter for Mailtrap Sending and Sandbox APIs.

API docs: https://api-docs.mailtrap.io/docs/mailtrap-api-docs/5tjdeg9545058-mailtrap-api


princemaple:
Thanks for the PR. It looks great 😃 I'll probably merge it in a couple of days, when I get the time to go through it in depth.

PR #828, opened 2023-10-25
Add a note about the Tailwindcss cdn when using a CSP

Hermanverschooten:
I am using a CSP in my router and noticed that the mailbox preview was missing all its styling.
I had to adapt my CSP to allow the use of the cdn.


princemaple:
💜

PR #830, opened 2023-11-10
Add Swoosh.ApiClient.Req

matthewlehner:
This adds an ApiClient based on [req](https://hex.pm/packages/req)

I based this on the Finch ApiClient as Req is built on top of Finch, it just has less configuration needed (it sets up its own supervised process).


princemaple (approved):
Looking good. Req is indeed a good to have default.

@josevalim @wojtekmach care to have a look?

matthewlehner:
I should also say, I've tested this with Mailgun and it works! ;)

I don't have access to any other providers at the moment or I would've tried them as well.

josevalim:
If we have Finch, I don't think we need Req, since Req is built on top of Finch. @wojtekmach may have a different opinion though.

princemaple:
Gonna merge this so people can take advantage of the more powerful and flexible options of `Req` on top of the existing benefits from using `Finch`. I think the options of `Req` is where the real value is, e.g. `retry` would be something a lot of users want to use.

mix.exs
```diff
@@ -68,6 +69,7 @@ defmodule Swoosh.Mixfile do
       {:telemetry, "~> 0.4.2 or ~> 1.0"},
       {:hackney, "~> 1.9", optional: true},
       {:finch, "~> 0.6", optional: true},
+      {:req, "~> 0.4", optional: true},
```
josevalim, line 72: So let’s add `or ~> 1.0` here too, as it is coming out soon. :)

princemaple: Thanks!

lib/swoosh/api_client/req.ex
```diff
@@ -0,0 +1,66 @@
...
+
+    options = email.private[:client_options] || []
+
+    # Set `decode_body` option to false as Swoosh's Mailgun adapter expects
+    # that the body is a binary. Req will run `Jason.decode` itself and
+    # `response.body` will already be a map if it is not set.
+    required_options = [
+      headers: [@user_agent | headers],
+      body: body,
+      decode_body: false
```
wojtekmach, line 53: Blast from the past! Similarly to `decode_body: false`, you might consider setting `retry: false`. Req is retrying by default, though I'm considering changing that before v1.0, slightly leaning towards. And so it wouldn't hurt to be super explicit here.

In swoosh README we have:

> One of the downsides of sending email asynchronously is that failures won't be reported to the user, who won't have an opportunity to try again immediately, and tasks by default do not retry on errors.

So yeah this is not totally accurate with regards to Req right and it can be indeed surprising to folks, especially if they end up putting stuff to Oban.

Apologies I haven't mentioned that sooner. Thanks for adding built-in support for Req.

josevalim: Shouldn't we have `retry: true` instead? :) I guess people can configure it as necessary too, since options are merged below, so perhaps no changes necessary here.

wojtekmach: Doing `retry: true` here in swoosh would be super explicit and protect against any changes in req in this area if that’s what you meant.

wojtekmach: Sorry, the current default is retry: :safe_transient. Safe meaning only GET/HEAD. So this is moot, I assume sending emails would be using POST. Sorry for noise!

PR #834, opened 2023-11-24
Do not send subject to customer.io when empty

caioaao:
Customer.io transactional emails can have subjects defined in the template, and if the subject is empty they won't deliver the email.

This PR changes the code so that, if the subject is an empty string, it won't set it in the request to customer.io. Previously it would only check if subject is `nil`, but that's not supported by swoosh's API (I still left the `nil` implementation in case someone managed to use it).

Also added a test checking that we're not sending empty `from` values, as those can also be populated by customer.io


caioaao:
Oh, forgot to mention: another approach would be to allow `subject` to be `nil`, but I didn't want to change swoosh's core API to support this

princemaple:
💜 release will come a bit later.

princemaple:
Released in 1.14.2

PR #837, opened 2023-12-13
Implement Proton SMTP Submission Adapter

mandax:
This PR adds an adapter for Proton SMTP Submission leveraging the SMTP Submission token connecting directly to Proton SMTP, not requiring ProtonBridge to be running as used in the ProtonBridge adapter.

The Proton token can be generated from ProtonMail -> Settings -> IMAP/SMTP:

proton-token-generator


princemaple:
This is no different from SMTP. I don't think this needs to be another adapter. I accepted the bridge one because it's special. I'll move your config to the docs part of PhotonBridge to show how it's normally configured without a bridge.

PR #842, opened 2024-01-08
fix comma issues on adapter config samples

SirWerto:
Changes:
- fix comma issues on adapter config samples (Scaleway and Dyn)


princemaple:
Thank you! 💜

PR #846, opened 2024-01-17
AmazonSES: add :ses_source option to set Source API parameter

adamu:
The SES API includes a [`Source` parameter](https://docs.aws.amazon.com/ses/latest/APIReference/API_SendRawEmail.html) than can be used to specify an email address that SES will use for bounce reports (equivalent to the `MAIL FROM` address when using SMTP).

This PR adds an optional `:ses_source` config option to the `AmazonSES` adapter (and by extension to the the `ExAwsAmazonSES` adapter). Providing the option sets the `Source` field in the API requests to the specified value.

---

Side note: I think it makes sense to have the adapter-specific API fields as options on the adapter, rather than adding them to the email. Even the existing `put_provider_option`s might be better as a adapter options, which can either be configured globally or passed in to `deliver`. Anyway, I need to set `Source`, so that's what this PR does 🙂


princemaple:
Looks reasonable. Thanks 👍

Issue #849, opened 2024-01-24
Multiple reply_to addresses for Mailgun


ghostdsb:
**Is your feature request related to a problem? Please describe.**
There should be a way to specify multiple addresses for reply_to.

**Is your feature request related to a specific adapter? Please specify.**
Mailgun

**Describe the solution you'd like**
As per discussions in this [issue](https://github.com/swoosh/swoosh/issues/738), adding `[mailbox()]` to reply_to type and handling `is_list` in adapter should do the trick

**Describe alternatives you've considered**
I considered setting the header value, but then we would have Reply-to(from default value) and Reply-To(from header) 

![swoosh](https://github.com/swoosh/swoosh/assets/20119474/dcda8727-b198-4a2b-a099-155019bc0536)

I do have a working PR ready that I used in a project, sending it for further discussion.

ghostdsb:
PR link mentioned above https://github.com/swoosh/swoosh/pull/850

princemaple:
Where does the default value come from? Is it from your application code? Adaptor code doesn't add anything if it's nil.

The pr looks pretty good though. I'll take a deeper dive soon.

ghostdsb:
Yes, you are right. I was confused by seeing `Reply-to` and `Reply-To` in the preview. I see that a `Reply-To` header does not overwrite `Reply-to`.

PR #850, opened 2024-01-24
Feat mailgun multiple reply to

ghostdsb:
feature addition for https://github.com/swoosh/swoosh/issues/849

Added `[mailbox]` to `reply_to` types and handle `is_list` recipients in `emails.ex` and mailgun adapter

swoosh-feat

now we are getting consistent reply-to


princemaple:
I simplified `reply_to` to only overwrite. Please take a look and let me know if you are happy for this to go in as it is now.

ghostdsb:
This looks good.

PR #853, opened 2024-01-25
Support multiple reply_to in sendgrid

### LINKED ISSUE #738: Multiple reply_to addresses (Sendgrid)
There should be a way to specify multiple addresses for `reply_to`.

**Is your feature request related to a specific adapter? Please specify.**
Sendgrid

Now there is a way to add multiple addresses in `reply_to` (at least in sendgrid) - https://docs.sendgrid.com/api-reference/mail-send/mail-send#multiple-reply-to-emails

I am thinking about preparing a PR but looks like it will affect `Email` type (which has `reply_to: mailbox() | nil`) so I am thinking how to do it and to not break everything else. Do you think it makes sense to change type to `[mailbox()] | []` and handle it everywhere?

princemaple:
blocked by #850 

close #738

### DIFF
--- a/lib/swoosh/adapters/sendgrid.ex
+++ b/lib/swoosh/adapters/sendgrid.ex
@@ -242,8 +242,11 @@ defmodule Swoosh.Adapters.Sendgrid do
 
   defp prepare_reply_to(body, %{reply_to: nil}), do: body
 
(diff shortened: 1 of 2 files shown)


princemaple:
FYI @ssbb

PR #859, opened 2024-02-06
clarify Postmark docs

TheArrowsmith:
I was trying to send an email with a template that doesn't have any interpolated variables, but I kept getting an error when trying to send it. Eventually I realised that I needed to pass an empty map as the `:template_model`. This PR updates the docs to make this clearer.


princemaple:
Thanks!

PR #870, opened 2024-02-27
add Swoosh.Adapters.Mua

### LINKED ISSUE #869: alternative SMTP adapter
👋

Just checking if there is any interest in an alternative SMTP adapter which doesn't use `gen_smtp`: http://github.com/ruslandoga/swoosh_mua

Instead of `gen_smtp` it uses a minimal Mint-like SMTP client http://github.com/ruslandoga/mua

I believe some of the TLS issues in OTP-26 would be resolved that way.

ruslandoga:
closes #869 

Is there a quick and easy way to run the SMTP integration tests?


ruslandoga:
I'll need to make Mua support Elixir v1.13 to make the tests pass.

princemaple:
With all the env vars set, run `mix test --include=integrations ./path/to/mua-integration-test`

ruslandoga:
Right, but what are those env vars set to? What kind of service do you use to run these tests? I used MailHog previously but it doesn't have STARTTLS support so the tests wouldn't pass.

princemaple:
Oh I thought you wanted to test it locally against your own gmail or something. These tests are supposed to be run against real services.

ruslandoga:
How would sending from `swoosh+smtp@#{config[:domain]}` work with gmail?

princemaple:
> I guess I'd need to modify the test to allow for custom sender address, right? I'll do it locally to run the tests but won't commit to avoid breaking anything.

That's right. 

Thanks!

princemaple:
Thanks for the great commit(s)ment.

It's late in Australia now and I'll get back to merge and release this tomorrow. Promise.

If you haven't already, might want to update the table in readme to include this one, also mark it as an SMTP alternative in the last column - people don't necessarily know what Mua is.

ruslandoga:
Screenshot 2024-02-27 at 20 22 04

But I'm not seeing any attachments. BTW, Swoosh.Adapter.SMTP adapter tests don't pass using this setup with the following errors:

```console
$ mix test --include integration test/integration/adapters/smtp_test.exs
Including tags: [:integration]

  1) test using Elixir.Swoosh.Adapters.SMTP simple deliver (Swoosh.Integration.Adapters.SMTPTest)
     test/integration/adapters/smtp_test.exs:37
     match (=) failed
     code:  assert {:ok, _response} = Swoosh.Adapters.SMTP.deliver(email, config)
     left:  {:ok, _response}
     right: {:error, {:retries_exceeded, {:network_failure, ~c"142.251.12.108", {:error, :timeout}}}}
     stacktrace:
       test/integration/adapters/smtp_test.exs:49: (test)

..

  2) test using Elixir.Swoosh.Adapters.SMTP deliver with attachment in memory (Swoosh.Integration.Adapters.SMTPTest)
     test/integration/adapters/smtp_test.exs:52
     match (=) failed
     code:  assert {:ok, _response} = Swoosh.Adapters.SMTP.deliver(email, config)
     left:  {:ok, _response}
     right: {:error, {:retries_exceeded, {:network_failure, ~c"142.251.12.108", {:error, :timeout}}}}
     stacktrace:
       test/integration/adapters/smtp_test.exs:67: (test)


Finished in 27.0 seconds (27.0s async, 0.00s sync)
4 tests, 2 failures

Randomized with seed 156324
```

princemaple:
image
I did some brief manual testing with Mua, seems to work OK now. The first email was pure text with attachment. The second was mixed and with attachment.

Was it fixed by the content length change?

princemaple:
Feel free to ignore the SMTP issue for now.

Do we have a scenario in Swoosh where `relay` is not present? Do we want to enforce it with `validate_config`?

We probably should validate `Mail` exists and a dependency.

See https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/mailgun.ex#L80 for an example of validating dependencies and configs.

Pleas take a look at https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/smtp/helpers.ex#L56-L57, do we want to do the same in Mua https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/mua.ex#L120 ?

If you are happy with the Mua side of things, I'll release v1.16.0

ruslandoga:
> Do we have a scenario in Swoosh where relay is not present?

Yes, this was the original idea behind Mua as all the other SMTP clients I could find required something like a relay. Mua can be used to send emails directly to the recipients (provided outgoing connections to port 25 are open and proper DNS records are set on the sender domain). I'll document it.

> Pleas take a look at https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/smtp/helpers.ex#L56-L57, do we want to do the same in Mua https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/mua.ex#L120 ?

Not sure, I'm not very familiar with RFC but in my understanding BCC was always hidden from the message body and only used for the RCPT TO commands.

princemaple:
Thanks for the detailed reply! I'll be in air / airports most of the time on 29th (Melbourne -> Frankfurt). My apology in advance for any inevitable delay.

princemaple:
> Hm, not sure. Right now Mail just drops them [in its renderer.](https://github.com/DockYard/elixir-mail/blob/d83c47f564372d0343a51b74425a627c8feddd52/lib/mail/renderers/rfc_2822.ex#L21)

This looks good enough. No need to explicitly drop it then.

PR #871, opened 2024-03-01
mua: validate deps and more docs

ruslandoga:
Continues #870


ruslandoga:
👋 @princemaple 

I think this is now ready for your review. I can't seem to be able to request it via GitHub UI.

princemaple:
Thank you very much @ruslandoga! This will go out in a release soon-ish.

PR #879, opened 2024-03-15
Fix Req header handling

wojtekmach:
Since Req v0.4.0 (*), request and response headers are _maps_ of name and lists of values, e.g.:

```elixir
%{"content-type" => ["text/html"]}
```

(*) there's a way to keep the old behaviour, setting `config :req, legacy_headers_as_lists: true` but at this point people should upgrade away from it so I wouldn't even worry about it.


princemaple:
Thanks!

PR #822, opened 2023-10-12
Update the UI for the mailbox viewer

dsincl12:
This will give the mailbox viewer a new UI.

Screenshot 2023-10-12 at 3 32 20 PM

Screenshot 2023-10-12 at 3 32 35 PM

Screenshot 2023-10-12 at 3 32 49 PM

PR #882, opened 2024-03-18
Mailbox Preview: more space for the HTML preview

justincy:
Follow up to #822 to make the HTML preview fill the available space and to condense the metadata table at the top so that we get even more space.

## Before

image

## After

image


princemaple:
Thanks!

PR #883, opened 2024-03-18
Fix Mailgun adapter incompatibility with Finch

### LINKED ISSUE #881: Mailgun provider is not compatible with Finch API client
### Describe the bug

If you replace hackney with finch the Mailgun provider will stop sending the request body. Here are two requests logged, the one with empty body is the finch and other one is hackney:
Screenshot 2024-03-18 at 10 36 42

My best guess is that Finch sees the `Content-Length: 0` request header and either discards the body or the end server just doesn't read it because of the header, while Hackney will actually ignore the header you would set manually and re-calculate it. The header is set for both API clients:
```
hackney_headers = [
   {"User-Agent", "swoosh/1.16.1"},
   {"Content-Type", "application/x-www-form-urlencoded"},
   {"Content-Length", "0"},
   {"User-Agent", "swoosh/1.16.1"},
   {"Accept", "*/*"},
   {"Authorization",
    "Basic REDACTED"}
 ]

finch_headers = [
    {"User-Agent", "swoosh/1.16.1"},
    {"Content-Type", "application/x-www-form-urlencoded"},
    {"Content-Length", "0"},
    {"User-Agent", "swoosh/1.16.1"},
    {"Accept", "*/*"},
    {"Authorization",
     "Basic REDACTED"}
]
```
and HTTP bin received a proper value in case of Hackney.

add `config :swoosh, :api_client, Swoosh.ApiClient.Finch` to the config

### Expected behavior

_No response_

### Your reproducible example

_No response_

### Screenshots or Videos

_No response_

### Platform

- OTP:
- Elixir:

### Additional context

_No response_

AndrewDryga:
Closes #881

### DIFF
--- a/lib/swoosh/adapters/mailgun.ex
+++ b/lib/swoosh/adapters/mailgun.ex
@@ -248,7 +248,8 @@ defmodule Swoosh.Adapters.Mailgun do
 
   defp prepare_template_options(body, _), do: body
 
-  defp encode_body(%{attachments: attachments} = params) do
+  @doc false
+  def encode_body(%{attachments: attachments} = params) do
     attachments
     |> Enum.reduce(
       params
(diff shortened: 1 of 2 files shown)


lib/swoosh/adapters/mailgun.ex
```diff
@@ -248,7 +248,8 @@ defmodule Swoosh.Adapters.Mailgun do
 
   defp prepare_template_options(body, _), do: body
 
-  defp encode_body(%{attachments: attachments} = params) do
+  @doc false
```
AndrewDryga, line 251: I had to make it public because `bypass` will receive the request built by Hackney so the new assertions will always pass, so we need an extra test to make sure the value is at least built properly.

PR #895, opened 2024-04-19
Add support for allow_nil_from

bernardd:
Mandrill email templates allow specification of the `from` email address and name on the template itself. Swoosh, however, requires `from` to be explicitly set which will cause the template values to always be overridden. This change adds the `allow_nil_from` config option which skips that check and allows use of the template-sourced values.

### DIFF
--- a/lib/swoosh/adapters/mandrill.ex
+++ b/lib/swoosh/adapters/mandrill.ex
@@ -78,6 +78,14 @@ defmodule Swoosh.Adapters.Mandrill do
 
     * `:tags` (list[string]) - a list of strings to tag the message with
 
+  ## Template-configured 'from' address
+
+  Mandrill templates allow you to configure the 'from' address in the template itself.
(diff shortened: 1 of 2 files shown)


princemaple:
Hi Bernard, thanks for the PR. I would prefer not adding a new config, as this would add complexity to understanding which adapter could use which configs.

There is a similar scenario in Brevo. I would like to suggest a similar solution. Simply documenting a special value `from`, e.g. `{:from, "TEMPLATE"}`, and have a special clause of `prepare from` that handles it.
https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/brevo.ex#L118

bernardd:
Makes sense to me. I've pushed a new version with that change.

princemaple:
Thanks Bernard. I'm at the airport. I'll get back to this later today, or tomorrow. Feel free to ping me if I forgot to do so. Cheers!

bernardd:
No rush, thanks @princemaple :)

PR #897, opened 2024-04-29
Add config options for AmazonSES adapter

otlaitil:
Implements support for `SourceArn`, `FromArn` and `ReturnPathArn` [request parameters](https://docs.aws.amazon.com/ses/latest/APIReference/API_SendRawEmail.html#API_SendRawEmail_RequestParameters) in the AmazonSES adapter. 

These parameters are necessary for scenarios such as sending emails across different AWS accounts where the sender identity is managed in a separate account (which is the use case I'm facing now).

I used this PR as a reference: https://github.com/swoosh/swoosh/pull/846 where the `Source` parameter was set as a config option. Other SES API request parameters `ConfigurationSetName` and `Tags` are passed as provider options. For consistency it might be better to use provider options for all of the request parameters. Maybe that could be another PR if we decide so?

### DIFF
--- a/lib/swoosh/adapters/amazon_ses.ex
+++ b/lib/swoosh/adapters/amazon_ses.ex
@@ -41,9 +41,14 @@ defmodule Swoosh.Adapters.AmazonSES do
 
   ### Optional
 
-  * `:ses_source` - An email address used for bounce reports. Will be set as the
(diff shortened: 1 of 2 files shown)


princemaple (approved):
Looks good.

princemaple:
Thanks @otlaitil !

princemaple:
Out in 1.16.5

otlaitil:
Thanks @princemaple !

PR #901, opened 2024-05-08
Specify the correct content_disposition and content_id

Hermanverschooten:
I recently migrated a project from using `Swoosh.Adapters.SMTP` to `Swoosh.Adapters.Mua`
and noticed that my inline attachments were no longer working.

Checking the code I saw that the `:type` and `:cid` fields were completely ignored when
creating the `Mail.Attachment`, hence this PR.

### DIFF
--- a/lib/swoosh/adapters/mua.ex
+++ b/lib/swoosh/adapters/mua.ex
@@ -185,7 +185,7 @@ defmodule Swoosh.Adapters.Mua do
     |> Enum.uniq()
   end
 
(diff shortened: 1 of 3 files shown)


Hermanverschooten:
I am currently closing this PR, as there is more problems with the `Mua` adapter when trying to use inline attachments.
It is missing the necessary headers. I'll open an issue for this.

Hermanverschooten:
I reopened and added a test

PR #905, opened 2024-05-14
add Swoosh.Adapters.ZeptoMail

gBillal:
Add adapter for Zoho's [`ZeptoMail`](https://www.zoho.com/zeptomail/)


lib/swoosh/adapters/zepto_mail.ex
```diff
@@ -0,0 +1,277 @@
...
+
+  defguardp is_template(email)
+            when is_map_key(email.provider_options, :template_key) or
+                   is_map_key(email.provider_options, :template_alias)
+
+  def deliver(%Email{} = email, config \\ []) do
+    do_deliver(:single, email, config)
+  end
+
+  def deliver_many([email], config \\ []) do
```
princemaple, line 106: This only matches a list of one item. Probably not what you intended.

If intended, we'd rather not implement this function.

princemaple (changes requested):
Please also update the readme adaptors list ;)

lib/swoosh/adapters/zepto_mail.ex
gBillal: I explained the reason behind this on the module documentation.
Sending batch email is a useful feature of ZeptoMail.  it works by sending one email to each recipient with an ability to customize variables for each one via `merge_info`.

If you think it's not how `deliver_many/2` is intended to be used, I'll try to make it as another option in the `config` field.

princemaple: I'll take a better look later. Cheers for the quick reply.

princemaple: Yes I think using the provider option would be more appropriate, instead of `deliver_many/2`.

Same templating functionality exists in many other providers.

PR #911, opened 2024-05-29
update mua to v0.2.0

ruslandoga:
This PR updates Mua docs for v0.2.0. The only real [change](https://github.com/ruslandoga/mua/pull/44) in v0.2.0 is splitting `transport_opts` into `tcp` and `ssl` specific ones since `:gen_tcp.connect` complains when it receives opts for `:ssl.connect`.


princemaple:
Hi, is it backwards compatible? Is Mua still able to handle the old configs if provided?

ruslandoga:
Hi! No, it's not backwards compatible, hence the version bump.

princemaple:
Cheers.

PR #918, opened 2024-06-05
fix broken anchor link in docs

jklq:
I noticed one broken link when reading the docs. I didn't find the same bug anywhere else.


princemaple:
Hi, this is due to having to work both on GitHub on Hex docs. I fixed it in https://github.com/swoosh/swoosh/commit/5f70eb27194bcd80f9b13c9ffb9672e004b91885. Closing now. Thanks for bringing it up!

Issue #854, opened 2024-01-29
`assert_no_email_sent` and `refute_email_sent` do not work with `deliver_many`


martinmaillard:
### Describe the bug

When using `deliver_many`, `assert_no_email_sent` and `refute_email_sent` do not raise `AssertionError`

### Steps to Reproduce the Bug or Issue

Here are test cases that can be added to `Swoosh.TestAssertionsTest`:

```elixir
    test "assert no email sent" do
      receive do
        _ -> nil
      end

      assert_no_email_sent()
    end

    test "assert no email sent with email sent" do
      assert_raise ExUnit.AssertionError, fn ->
        assert_no_email_sent()
      end
    end

    test "refute email sent" do
      receive do
        _ -> nil
      end

      refute_email_sent()
    end

    test "refute email sent with email sent" do
      assert_raise ExUnit.AssertionError, fn ->
        refute_email_sent()
      end
    end
```

### Platform

- OTP:
- Elixir:

PR #924, opened 2024-06-18
Complete Swoosh.X.TestAssertions

### LINKED ISSUE #909: Finish experimental assertions module
Bring `Swoosh.X.TestAssertions` on par with `Swoosh.TestAssertions`

edgarlatorre:
Move the methods `assert_emails_sent`, `assert_email_not_set` and `assert_no_email_sent` from `Swoosh.TestAssertions` to `Swoosh.X.TestAssertions`. I also moved the private methods used by the ones I moved, I was wondering if we should remove all those assert_equals and do the assertions inside the assert_emails_sent. Finally I added the tests to ensure those methods are working as they should.

closes #909


lib/swoosh/x_test_assertions.ex
```diff
@@ -275,6 +295,96 @@ defmodule Swoosh.X.TestAssertions do
...
+    assert length(emails) == length(params_map_list)
+
+    emails
+    |> Enum.zip(params_map_list)
+    |> Enum.each(fn {email, params_map} ->
+      Enum.each(params_map, &assert_equal(email, &1))
+    end)
+  end
+
+  defp assert_equal(email, {:subject, %Regex{} = value}),
```
edgarlatorre, line 343: Should we do the assertions below directly in the assert_emails_sent? I think it would be more explicity what we are testing.

princemaple: You can probably reuse these https://github.com/swoosh/swoosh/blob/f21726a81f7146a99b16fd43c929e6e0d3ee48bf/lib/swoosh/x_test_assertions.ex#L169-L209 and do similar things to https://github.com/swoosh/swoosh/blob/f21726a81f7146a99b16fd43c929e6e0d3ee48bf/lib/swoosh/x_test_assertions.ex#L123-L166. When I said "The existing implementation can serve as a good implementation guide" I meant the X version. Sorry I wasn't clear. 😢

princemaple: I think `has?` is missing a couple regex cases, comparing to `assert_equal`

edgarlatorre: Sorry for the confusion :smile: 

I removed the assert_equal in order to use `has?` . Also , I implemented the regex case and a couple of tests for subject which was the only case I noticed comparing with assert_equal. Finally, I changed the implementation of the `assert_emails_sent`. Let me know what you think.

princemaple:
Thanks! Give me some time to think about whether/how I can incorporate https://github.com/swoosh/swoosh/issues/854 into this PR, or as a follow up.

princemaple:
@jakub-gonet care to take a look to see if there is anything to add?

lib/swoosh/x_test_assertions.ex
```diff
@@ -64,6 +66,24 @@ defmodule Swoosh.X.TestAssertions do
...
+  @doc ~S"""
+  Asserts `email` was not sent.
+
+  Performs exact matching of the email struct.
+  """
+  @spec assert_email_not_sent(Email.t()) :: boolean | no_return
+  def assert_email_not_sent(%Email{} = email) do
+    refute_email_sent(email)
+  end
+
```
princemaple, line 86: note to myself: might still want to remove these to simply favor the refute functions.

PR #934, opened 2024-07-13
mua: no mx when relay

ruslandoga:
This PR removes MX lookups when relays are used.

Related issue: https://github.com/plausible/analytics/discussions/4342


mix.exs
```diff
@@ -76,7 +76,7 @@ defmodule Swoosh.Mixfile do
       {:req, "~> 0.5 or ~> 1.0", optional: true},
       {:mail, "~> 0.2", optional: true},
       {:gen_smtp, "~> 0.13 or ~> 1.0", optional: true},
-      {:mua, "~> 0.2.0", optional: true},
+      {:mua, "~> 0.2.3", optional: true},
```
ruslandoga, line 79: Pinning to at least 0.2.3 for https://github.com/ruslandoga/mua/pull/55

PR #935, opened 2024-07-13
mua: update docs

ruslandoga:
This PR updates docs to describe the latest defaults for handling CA certs.


lib/swoosh/adapters/mua.ex
```diff
@@ -1,91 +1,104 @@
...
+  > Each `deliver/2` call results in a new connection to the recipient's email server.
 
-  ## Sending email via a relay
+  ## Sending Email via a Relay
 
-  When `relay` option is set, this adapter will send email through that relay.
-  The relay would usually require authentication. For example, you can use your own
-  GMail account with an app password.
+  When the relay option is set, emails are sent through the specified relay, typically requiring authentication.
+  For example, you can use your Gmail account with an app password.
```
ruslandoga, line 63: Would it be OK to create a guide like https://gist.github.com/ruslandoga/c94ce526231fb77930132aaeda3fc3c9 and link to it from here or would it be overkill?

princemaple: I'm happy to merge in any docs related improvements. There cannot be too much documentation... or could it? ;)

PR #938, opened 2024-07-21
mua: cleanup test

ruslandoga:
~~I'll create a Google App password and rerun the integration test later today.~~

I used [Mailpit](https://github.com/axllent/mailpit) instead.

### DIFF
--- a/test/integration/adapters/smtp_test.exs
+++ b/test/integration/adapters/smtp_test.exs
@@ -16,15 +16,16 @@ defmodule Swoosh.Integration.Adapters.SMTPTest do
     ]
   end
 
-  defp config(Swoosh.Adapters.Muaua) do
-    [
-      relay: System.get_env("SMTP_RELAY"),
-      domain: System.get_env("SMTP_DOMAIN"),
-      auth: [
-        username: System.get_env("SMTP_USERNAME"),
-        password: System.get_env("SMTP_PASSWORD")
-      ]
-    ]
+  defp config(Swoosh.Adapters.Mua) do
+    relay = System.get_env("SMTP_RELAY", "localhost")
+    domain = System.get_env("SMTP_DOMAIN", "mua.local")
+    port = String.to_integer(System.get_env("SMTP_PORT", "1025"))
+
+    username = System.get_env("SMTP_USERNAME")
+    password = System.get_env("SMTP_PASSWORD")
+    auth = if username && password, do: [username: username, password: password]
+
+    [relay: relay, port: port, domain: domain, auth: auth]
   end
 
   for adapter <- [Swoosh.Adapters.SMTP, Swoosh.Adapters.Mua] do


test/integration/adapters/smtp_test.exs
```diff
@@ -16,15 +16,25 @@ defmodule Swoosh.Integration.Adapters.SMTPTest do
     ]
   end
 
-  defp config(Swoosh.Adapters.Muaua) do
```
ruslandoga, line 19: Muaua -- not sure how it was working before.

test/integration/adapters/smtp_test.exs
```diff
@@ -16,15 +16,25 @@ defmodule Swoosh.Integration.Adapters.SMTPTest do
...
+    username = System.get_env("SMTP_USERNAME")
+    password = System.get_env("SMTP_PASSWORD")
+
+    auth =
+      cond do
+        username && password ->
+          [username: username, password: password]
+
+        username || password ->
+          raise ArgumentError, "both SMTP_USERNAME and SMTP_PASSWORD need to be provided"
```
ruslandoga, line 30: Note to self: move this check to mua.

test/integration/adapters/smtp_test.exs
```diff
@@ -16,15 +16,16 @@ defmodule Swoosh.Integration.Adapters.SMTPTest do
...
-      domain: System.get_env("SMTP_DOMAIN"),
-      auth: [
-        username: System.get_env("SMTP_USERNAME"),
-        password: System.get_env("SMTP_PASSWORD")
-      ]
-    ]
+  defp config(Swoosh.Adapters.Mua) do
+    relay = System.get_env("SMTP_RELAY", "localhost")
+    domain = System.get_env("SMTP_DOMAIN", "mua.local")
+    port = String.to_integer(System.get_env("SMTP_PORT", "1025"))
```
ruslandoga, line 22: I'm using env vars that work with Mailpit by default.

```sh
$ docker run -d --rm -p 1025:1025 -p 8025:8025 -e "MP_SMTP_AUTH_ACCEPT_ANY=1" -e "MP_SMTP_AUTH_ALLOW_INSECURE=1" --name mailpit axllent/mailpit
```

Screenshot 2024-12-11 at 15 59 59

Screenshot 2024-12-11 at 15 59 55

I think we can also add it to CI unless there is already something like that. Here's how I use it in Mua:
- https://github.com/ruslandoga/mua/blob/99e905252201729e3d38ae643bce31977c550b0a/test/test_helper.exs#L15-L27
- https://github.com/ruslandoga/mua/blob/99e905252201729e3d38ae643bce31977c550b0a/.github/workflows/mix.yml#L15-L28

princemaple: Sounds good. We can do it in a later PR.

Issue #578, opened 2021-03-08
Swoosh breaking for major version detection for gen_smtp 1.1.1


aselder:
Today we updated to gen_smtp version 1.1.1 and started getting errors like:

`(BadMapError expected a map, got: [])`

stack trace:
```
BadMapError: expected a map, got: []
  File "maps.erl", line 188, in :maps.get/3
  File "/app/deps/gen_smtp/src/mimemail.erl", line 708, in :mimemail.ensure_content_headers/7
  File "/app/deps/gen_smtp/src/mimemail.erl", line 160, in :mimemail.encode/2
  File "lib/api_v1/email/adapters/amazon_ses.ex", line 14, in ApiV1.Email.Adapters.AmazonSes.deliver/2
```

It looks like a change in gen_stmp broke the major version detection in Swoosh. See https://github.com/gen-smtp/gen_smtp/compare/1.1.0...1.1.1

It looks like they renamed the application from `gen_smtp_application` to `application`, which swoosh was relying on.

See https://github.com/swoosh/swoosh/blob/0ea43246d7fea0ae78fb528121be7b6fd87efc5a/lib/swoosh/adapters/smtp/helpers.ex#L23

princemaple:
Thanks for the helpful information.

princemaple:
Fixed in 1.3.4

CGenie:
I'm using swoosh 1.5.2 and getting the same error as in the description of the ticket with the simple code:
```elixir
email = new |> from("<some-email>") |> to("<some-other-email>") |> subject("Test") |> text_body("Hello world")
WODR.Importer.Mailer.deliver(email)
```

princemaple:
@CGenie try deleting the `_build` folder and recompile everything.

CGenie:
Ah ok seems to work, thank you!

Issue #552, opened 2020-11-24
gen_smtp 1.0 does not seem to be working 


gabrielrinaldi:
I tried to upgrade to the latest version of swoosh (1.1.1) and gen_smtp (1.0) and while it seems that swoosh now supports gen_smtp 1.0 I still get an error:
```
* (BadMapError) expected a map, got: [{"content-type-params", [{"charset", "utf-8"}]}, {"disposition", "inline"}, {"disposition-params", []}]
    (stdlib 3.13.2) maps.erl:188: :maps.get(:disposition, [{"content-type-params", [{"charset", "utf-8"}]}, {"disposition", "inline"}, {"disposition-params", []}], "inline")
    (gen_smtp 1.0.1) /tmp/build_d992ce7d/deps/gen_smtp/src/mimemail.erl:721: :mimemail.ensure_content_headers/7
    (gen_smtp 1.0.1) /tmp/build_d992ce7d/deps/gen_smtp/src/mimemail.erl:849: :mimemail.encode_component_part/1
    (gen_smtp 1.0.1) /tmp/build_d992ce7d/deps/gen_smtp/src/mimemail.erl:827: anonymous fn/2 in :mimemail.encode_component/5
    (stdlib 3.13.2) lists.erl:1254: :lists.flatmap/2
    (gen_smtp 1.0.1) /tmp/build_d992ce7d/deps/gen_smtp/src/mimemail.erl:824: :mimemail.encode_component/5
    (gen_smtp 1.0.1) /tmp/build_d992ce7d/deps/gen_smtp/src/mimemail.erl:163: :mimemail.encode/2
```

I know the breaking change on gen_smtp was switching from lists to maps, and I am not sure, but the issue might be on the `prepare_part` helper function.

Or of course, I am doing something wrong.

princemaple:
Hi, do you see anything about Swoosh in the stacktrace?
Could you try removing `_build` and recompile?

gabrielrinaldi:
@princemaple yes, the error is triggered from here:
`(swoosh 1.1.1) lib/swoosh/adapters/smtp.ex:46: Swoosh.Adapters.SMTP.deliver/2`

Removing `_build` does not change anything, unfortunately.

princemaple:
The parameter is conditionally compiled with the gen_smtp version.
https://github.com/swoosh/swoosh/blob/master/lib/swoosh/adapters/smtp/helpers.ex#L21-L29

I'm going to test with my gmail account...

[...]

PR #944, opened 2024-08-17
Explain how to recompile after installing gen_smtp

aj-foster:
Hi there 👋🏼 

I recently ran into the same issue as described in #552 and #578. The solution was to force a recompilation of Swoosh after installing `gen_smtp`. This PR adds a related note to the moduledoc for the adapter.


princemaple:
Thanks 👍

axelson:
Thank you for this, I just ran into it as well!

PR #946, opened 2024-08-19
Write docs for functions in Mailer __using__ macro 

### LINKED ISSUE #939: Missing documentation of deliver!/2
### Describe the bug

- [`deliver!/2`](https://github.com/swoosh/swoosh/blob/v1.16.10/lib/swoosh/mailer.ex#L132) is not documented
- only `deliver/2` is [documented](https://hexdocs.pm/swoosh/Swoosh.Mailer.html#deliver/2)

See the latest docs: https://hexdocs.pm/swoosh/Swoosh.Mailer.html#deliver/2

### Expected behavior

_No response_

### Your reproducible example

_No response_

### Screenshots or Videos

_No response_

### Platform

_No response_

### Additional context

_No response_

ivanhercaz:
Hi!
I found #939 and decide to check if I could make my first contribution, at least writing docs, so here here it is!

With this PR I write basic documentation for the functions inside the `__using__` `defmacro`:

- `deliver/2`
- `deliver!/2`
- `deliver_many/2`

I know that sometimes functions inside macro are not documented, however, I think it may be useful when the user generates the documentation for its project or even if the user has an editor that shows this docs (at least in my VSCode with ElixirLS in an test project doesn't shows the docs, but I imagine there is some way).

I can't say it fix #939, because I think the issue it is a bit in the wrong direction because `deliver!/2` won't be never in Swoosh documentation because it is implemented outside the `__using__`.

Well, hope this little contribution may be useful!

### DIFF
--- a/lib/swoosh/mailer.ex
+++ b/lib/swoosh/mailer.ex
@@ -150,6 +150,11 @@ defmodule Swoosh.Mailer do
         end
       end
 
+      @doc ~S"""
+      Delivers a list of emails with the given config.
+
+      It accepts a list of `%Swoosh.Email{}` as its first parameter.
+      """
       @spec deliver_many(list(%Swoosh.Email{}), Keyword.t()) :: {:ok, term} | {:error, term}
       def deliver_many(emails, config \\ [])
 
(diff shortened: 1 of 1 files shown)


princemaple:
> I can't say it fix #939, because I think the issue it is a bit in the wrong direction because `deliver!/2` won't be never in Swoosh documentation because it is implemented outside the `__using__`.

Yeah as far as I'm concerned this part is a `wont fix`.

princemaple:
Thanks. close #939

PR #949, opened 2024-08-25
Implement Postal adapter

onvlt:
Postal is open-source, self hosted transactional mail service. See https://docs.postalserver.io/

This is my first contribution so I'm open to feedback.


onvlt:
@princemaple Thanks for feedback, I've added the tests.

princemaple:
Thanks for the contribution! @onvlt

PR #955, opened 2024-09-13
Update the MSGraph Adapter Dependency

### LINKED ISSUE #954: MsGraph Adapter failing to find gen_smtp dependency
### Describe the bug

I’m attempting to utilize the new MSGraph adapter and it mostly works - the one issue I’m having is that even though I’ve got gen_smtp in my deps (and in the lock), I’m getting the error:

```
[error] The following dependencies are required to use Swoosh.Adapters.MsGraph:

- :gen_smtp
```

The error is triggered by this bit of code:

```
  @spec validate_dependency([module | {atom, module}]) ::
          :ok | {:error, [module | {:atom | module}]}
  def validate_dependency(required_deps) do
    if Enum.all?(required_deps, fn
         {_lib, module} -> Code.ensure_loaded?(module)
         module -> Code.ensure_loaded?(module)
       end),
       do: :ok,
       else: {:error, required_deps}
  end
```
 
Unless I go and manually update the required dependencies for that adapter to be [gen_smtp: :gen_smtp_client] as with other adapters that have any bit of similar dependency requirement, it fails with an error (with the change, everything seems to work fine).

So what I’m kinda looking for help here on is twofold:

1) Assuming the dependency definition is correct, how do I get it to pick up gen_smtp?

2) Is the dependency definition correct? It likely only requires the client, but I’m not super aware of the internals here of how this adapter is referencing it. Even further, does this requirement make any sense at all when using the MS Graph API, since it’s really just backed by a Rest API?

Thanks for any insight someone might have on either of these!

Brought the MsGraph adapter into an umbrella project, with gen_smtp as a dependency, and tried to use it.

### Expected behavior

_No response_

### Your reproducible example

_No response_

### Screenshots or Videos

_No response_

### Platform

- OTP: 26
- Elixir: elixir 1.16.1

### Additional context

_No response_

LetThereBeDwight:
Attempting to resolve #954 

This dependency is more generic than it needs to be _and_ seems to consistently fail in more recent builds as I can't seem to get gen_smtp to return true for Code.ensure_loaded at all, with any variations on how it's specified as a dependency.

### DIFF
--- a/lib/swoosh/adapters/ms_graph.ex
+++ b/lib/swoosh/adapters/ms_graph.ex
@@ -34,7 +34,7 @@ defmodule Swoosh.Adapters.MsGraph do
 
   use Swoosh.Adapter,
     required_config: [:auth],
-    required_deps: [:gen_smtp]
+    required_deps: [gen_smtp: :gen_smtp_client]
 
   require Logger
   alias Swoosh.Email


princemaple:
Oh yeah this looks like the right fix.

LetThereBeDwight:
@princemaple when will this get to hex?

princemaple:
When I get up ;) which is now

PR #956, opened 2024-09-16
Add CSS labels to email detail elements

alexslade:
This allows test assertions using a tool that drives the browser, or a browser-like interface. (E.g. Wallaby or PhoenixTest).

I follwed a rough BEM naming convention to avoid class or ID collisions.

I avoided naming the `headers` and `provider_options` as I can't guarantee these don't have repeated named values. (I believe headers are allowed to appear more than once, per the spec)

[RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322)

Here's a sample from our test suite that is made easier with the above addition. 

```
conn
    |> visit("/dev/mailbox")
    |> click_link("Here's my email")
    |> assert_has("#email-details__to .details__value", text: "alex@example.com")
```


princemaple:
Ummm why would you test this page? You could just test against the emails in memory directly. Am I missing a use case?

alexslade:
@princemaple 

> This allows test assertions using a tool that drives the browser, or a browser-like interface. (E.g. Wallaby or PhoenixTest).

These tools can't use the in-memory checks, as noted in the swoosh docs: https://hexdocs.pm/swoosh/Swoosh.TestAssertions.html 

It looks like the current supported workflow for these tests is to access the local mailbox, but the local mailbox doesn't have many semantic labels that can be used to identify values. This PR adds some.

princemaple:
I meant that you shouldn't be using those tools to test this page. This page itself is for dev only, and uses the emails in memory. In your projects you should be testing against the emails in memory directly. Unless you meant testing this page in Swoosh...?

alexslade:
@princemaple The swoosh docs say that this approach should be used, or have I misunderstood? 

> For feature/E2E tests, you should use Swoosh.Adapters.Local adapter. In your test, instead of calling assert_email_sent, you should navigate to the preview url with your E2E tool (e.g. wallaby) and test that the email is in the inbox.

princemaple:
Hmm I think I'll update that instruction. Even in E2E I'd check local mailbox with code directly. Instead of navigating to the page. The page is for human ;)

princemaple:
However, I think I will merge this one, too.

alexslade:
Thanks @princemaple. But can I ask why you would change the docs? 

> Note: [Swoosh.TestAssertions](https://hexdocs.pm/swoosh/Swoosh.TestAssertions.html) works for unit tests and basic integration tests. Unfortunately, it's not going to work for feature/E2E tests. The mechanism of assert_email_sent is based on messaging sending between processes, and is expecting the calling process (the one that calls assert_email_sent) to be the calling process of Mailer.deliver, or be the parent process of the whatever does the Mailer.deliver call.

> For feature/E2E tests, you should use [Swoosh.Adapters.Local](https://hexdocs.pm/swoosh/Swoosh.Adapters.Local.html) adapter. In your test, instead of calling assert_email_sent, you should navigate to the preview url with your E2E tool (e.g. wallaby) and test that the email is in the inbox.

Are you saying that `assert_email_sent` now works differently and doesn't need the calling process to be the same as the process of Mailer.deliver?

princemaple:
I'm simply noting that when testing within elixir you should check the local adapter storage directly. Navigating to the page would be the last resort if your e2e was run outside elixir.

PR #965, opened 2024-11-07
feat(mailjet): add event_payload to provider options

mrdotb:
I need this feature from mailjet that is not implemented.
https://dev.mailjet.com/email/guides/send-api-v31/#tag-email-messages
I added the possibility to add an event_payload


princemaple:
Cool.

PR #967, opened 2024-11-07
Handle direct URL specification on MSGraph.deliver

### LINKED ISSUE #966: Support Distribution Lists with the MSGraph Adapter
The MSGraph Adapter currently assumes that the from address is always directly tied to the URL used to post the message.

**Is your feature request related to a specific adapter? Please specify.**
MSGraph

**Describe the solution you'd like**
An optional, configurable URL that can be passed in on `deliver`, allowing a client to directly specify the URL that the message would be posted to.

**Describe alternatives you've considered**
Raw MSGraph API handling which...would be much harder as it would need to implement everything swoosh does + the above

**Additional context**
The driver here is allowing the adapter to handle Distribution Lists, though in the wider scheme allows swoosh clients to utilize this in ways to support varying forms of "send as" and "send on behalf of" via MSGraph API.

PR incoming to make this happen.

LetThereBeDwight:
Closes #966


princemaple:
An easy LGTM. Thanks!

LetThereBeDwight:
Can I expect a hex push in the nearish future @princemaple ?

princemaple:
It's already out ;)

PR #971, opened 2024-11-12
Sendgrid: Support Mail Body Compression

luhagel:
Heya,
Sendgrid supports sending in gzip compressed bodies. (see [here](https://www.twilio.com/docs/sendgrid/api-reference/mail-send/mail-send#mail-body-compression))

This PR adds an option for that to the Adapter.

I'm somewhat unsure how to properly test this without duplicating most of the existing testsuite, so any pointers would be welcome here :)

Also, if there is a less invasive way to set the headers/compress the body I'm happy to make adjustments as needed.


princemaple:
Since `bypass` is a proper plug server, I think we can simply test if the header is expected, and the content is correctly parsed by the server, when compress is set to `true`. We don't need to test every scenario in a compressed setting.

luhagel:
Gonna need some more time with this, we'll need a custom Plug Parser for, should find some time the next few days 👍

kevinam99:
@luhagel - are you still working on this? I have some bandwidth and I think I can help. I'll appreciate any earlier context you have on this :)

luhagel:
@kevinam99 Heya, 
planning to over the holidays - the last few weeks were pretty rough work-wise ^^

I'll gladly accept any help you can spare :)

kevinam99:
@luhagel okay cool, let me take a shot at this

kevinam99:
Okay, I think I have something here. Noticed a weird behaviour with the parsers https://github.com/luhagel/swoosh/pull/1

Had to add a check for the gzip parser in conn_parser based on the content-encoding header because when I used it in the regular list of parsers, the `parse` callback wouldn't get invoked at all leading the test to fail. If I add the Gzip parser at the top of the list, only this one test would pass

kevinam99:
Never mind. Managed to fix the problem :)

princemaple:
Hi all, turned out that a custom parser wasn't the correct way to go. I changed it to use a custom body reader. I think we are all good now.

princemaple:
Thank you both!

Issue #902, opened 2024-05-08
Mua adapter and inline attachments


Hermanverschooten:
### Describe the bug

When trying to add an attachment to be displayed inline, this is added as a regular attachment and as such unavailable.

### Steps to Reproduce the Bug or Issue

Create a mail with an attachment with `type: :inline` and deliver it through `Swoosh.Adapters.Mua`.

this is what is generated by `Swoosh.Adapters.Mua`

```
Content-Type: multipart/alternative; boundary="AD32D554819F6400EE6B7E05"
Message-Id: <20240508160918.3F6C720141@ferd-verbiest.localdomain>
Date: Wed,  8 May 2024 16:09:18 +0000 (UTC)

--AD32D554819F6400EE6B7E05
Content-Type: text/html
Content-Transfer-Encoding: quoted-printable

 html content....

--AD32D554819F6400EE6B7E05
Content-Type: image/png
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename=qr.png

base64-encoded image

--AD32D554819F6400EE6B7E05--
```

And this by `Swoosh.Adapters.SMTP`

```
--_=2f2o332v3p4x0k4z3q446j22563r0k39=_
Content-Type: multipart/related;
	boundary="_=474g4y642g4t0q1k615k0u202x3b1w5d=_"
Content-Disposition: inline

--_=474g4y642g4t0q1k615k0u202x3b1w5d=_
Content-Type: text/html;
	 charset="utf-8"
Content-Transfer-Encoding: quoted-printable
Content-Disposition: inline

 html content....

--_=474g4y642g4t0q1k615k0u202x3b1w5d=_
Content-Transfer-Encoding: base64
Content-Id: <qr.png>
Content-Type: image/png
Content-Disposition: inline;
	filename=qr.png

base64-encoded image

--_=474g4y642g4t0q1k615k0u202x3b1w5d=_--

--_=2f2o332v3p4x0k4z3q446j22563r0k39=_--
```

I tried to fix the attachment issue with a PR, but while this added the correct headers for the inline attachment. The whole mail got screwed up because of the invalid headers.

### Expected behavior

The mail should have the necessary headers and the attachment should be marked as inline.

### Platform

- OTP: 26.2.3
- Elixir: 1.16.2

Hermanverschooten:
Apparently there is a PR in `Mail` to address the headers issue, https://github.com/DockYard/elixir-mail/pull/137.
[...]

PR #979, opened 2024-12-12
mua: warn on inline attachments

ruslandoga:
Context: #902 

- [ ] add test


ruslandoga:
Closing in favor of the real fix: https://github.com/swoosh/swoosh/issues/902#issuecomment-2809668360

PR #981, opened 2024-12-14
Adapters: Add documentation about setting the base_url for ZeptoMail …

reimeri:
Hi,

I had some problems with the ZeptoMail adapter which were caused by the `base_url` being different than the one set by default in the adapter (default: https://api.zeptomail.com/v1.1). For me the `base_url` should have been https://api.zeptomail.eu/v1.1 I would guess because I am using ZeptoMail from EU. Looking at the adapter implementation I noticed that it was already possible to change the `base_url` but this was not documented in the ZeptoMail page so I added it to the documentation.


princemaple:
Thanks 🙏

PR #978, opened 2024-12-11
mua: fix default message-id

ruslandoga:
Context: https://github.com/swoosh/swoosh/pull/945#issuecomment-2532876675

PR #982, opened 2024-12-15
mua: fix mail.from

ruslandoga:
Fixes the error reported in https://github.com/swoosh/swoosh/pull/978#issuecomment-2543317353 and adds some basic tests using Mailpit.

And I think it might make sense to retire the Swoosh version with the broken Swoosh.Adapters.Mua:

```console
$ mix hex.retire swoosh 1.17.4 deprecated --message "broken Swoosh.Adapters.Mua"
```

🙏 Sorry for the trouble! 🙏

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /cover
 /deps
 /doc
+/tmp
 erl_crash.dump
 *.ez
 .elixir_ls/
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 /cover
 /deps
 /doc
+/tmp
```
ruslandoga, line 5: Created by `@tag :tmp_dir` which is used in the attachments test.

princemaple:
Thanks for the quick fix!

PR #991, opened 2025-01-10
Sort list of adapters in readme

leifmetcalf:
I was checking if Swoosh supports MailPace and saw the first few items were sorted, so I assumed the whole list was sorted and missed that Swoosh actually does support MailPace.


princemaple:
Sorry, this list is not meant to be sorted. It's kind of by popularity and some other factors.

PR #996, opened 2025-01-27
fix: add csp nonce

yordis:


princemaple:
👀

yordis:
Related to https://github.com/phoenixframework/phoenix_live_dashboard/pull/459

@josevalim @princemaple do you know why I can not use `img` or `script` nonce attribute? It is removing my value for some reason ...

yordis:
https://stackoverflow.com/questions/55670985/google-chrome-stripping-nonce-values-from-script-tags 


Fun .... 

I am guessing that the CSS is being injected by the cdn tailwind script? If so, we may need to consider an alternative if possible

yordis:
Open to feedback thus far

princemaple:
Could you first help me understand why we do this for the server that'd only ever be run on dev?

princemaple:
If actually required, we can run tailwind on the template and release the dist css file.

yordis:
> If required, we can run Tailwind on the template and release the CSS file in the dist folder.

Regardless of this pull request, that sounds like a good idea; I wasn't expecting it to use that JavaScript.

---

> Could you first help me understand why we do this for the server that will only ever be run in development?

It depends on what "development" means. I have Swoosh as part of an internal/admin routing setup with Oban and Phoenix Dashboard. We enabled CSP in such routing, so I thought, why not just make it work instead of disabling the CSP for the interface?

I don't have any strong opinion either way, but it feels like we're so close. People can deploy the "development" adapter to a development release environment where it's fine to use it, or local development doesn't need to worry about CSP working or not.

I copied most elements from the Phoenix Dashboard.

princemaple:
> Regardless of this pull request, that sounds like a good idea; I wasn't expecting it to use that JavaScript.

Right. See #998 

Let's make it work with CSP.

yordis:
done, I think

princemaple:
Thanks @yordis

PR #1003, opened 2025-02-05
Fix broken attachments on Scaleway adapter

olivermt:
This PR adds the required `type` prop (Scaleways name for "content type") to Scaleway provider.

Any request without "type" fails with 400 error after Scaleway updated their API lately.

Scaleway api docs here: https://www.scaleway.com/en/developers/api/transactional-email/#path-emails-send-an-email


princemaple:
Thanks!

PR #1012, opened 2025-02-27
Implement loops.so adapter

caioaao:
https://loops.so/

Loops transactional API is very limited, so most of the good stuff has to be left out, like settings body, CC, etc. This is all we have: https://loops.so/docs/api-reference/send-transactional-email


princemaple:
If the wrapper is so thin, I don't even know why someone would want to use swoosh for this. 🤔 

Is this a popular service?

caioaao:
> If the wrapper is so thin, I don't even know why someone would want to use swoosh for this. 🤔

Good question! Our use case here is we're migrating from another provider, but here's some stuff I could think of:
1. So we can swap it by swoosh's test/dev adapters during development
2. Helps new devs that are familiar with swoosh by providing this as a plug n play setup

> Is this a popular service?

Relatively, yes. It's fairly new (2yrs) but gaining traction. My best guess is it's as popular as brevo, although it's hard to find any good metric due to its _very_ common name 😅  https://www.producthunt.com/products/loops

On another note though, their API might be too limited. We can't even provide a `sender`, which breaks swoosh's assumptions apparently, since we can't `deliver` an email without a provided sender 🫠

princemaple:
> On another note though, their API might be too limited. We can't even provide a sender, which breaks swoosh's assumptions apparently, since we can't deliver an email without a provided sender 🫠

This is fine. I think I've used a placeholder sender somewhere in another adapter.

I reckon you convinced me to 89.5% 😉. I'll take another look either over the weekend or next week. Cheers!

lib/swoosh/adapters/loops.ex
```diff
@@ -0,0 +1,121 @@
+defmodule Swoosh.Adapters.Loops do
+  @moduledoc ~s"""
+  An adapter that sends email using the Loops API.
+
+  For reference: [CustomerIO API docs](https://loops.so/docs/api-reference/send-transactional-email)
```
princemaple, line 5: s/CustomerIO/Loops/

lib/swoosh/adapters/loops.ex
```diff
@@ -0,0 +1,121 @@
...
+      defmodule Sample.Mailer do
+        use Swoosh.Mailer, otp_app: :sample
+      end
+
+  ## Using with provider options
+
+      import Swoosh.Email
+
+      new()
+      |> to("katy@example.com")
```
princemaple, line 27: mention need a placeholder `from`? which then gets ignored.

princemaple: if we ever get to 2.0, I'd remove the requirement of `from`, maybe...

princemaple (changes requested):
And we need readme

princemaple:
Thanks ;)

PR #1017, opened 2025-03-06
Fix: Prevent zeptomail error when receiving non json response body on 500

atoncetti:
Currently Zeptomail returns "\n" as body on 500 error, this breaks `Swoosh.json_library().decode!(body)}` since it is not valid json.

The proposed fix tries to decode if possible, or returns the raw body otherwise.


princemaple:
Thank you!

PR #1031, opened 2025-04-10
make links clickable in text email preview

SteffenDE:
also: set target="_top" inside iframe links to redirect outside of the iframe.

This should provide a lot better dev UX when using the local preview.

### DIFF
--- a/lib/plug/templates/mailbox_viewer/index.html.eex
+++ b/lib/plug/templates/mailbox_viewer/index.html.eex
@@ -119,7 +119,7 @@
                 </a>
               </div>
               <div class="grow">
-                <iframe src="<%= to_absolute_url(@conn, @email.headers["Message-ID"]) %>/html" class="w-full mt-1 -mb-4 h-full" frameborder="0"></iframe>
+                <iframe id="html-mail" src="<%= to_absolute_url(@conn, @email.headers["Message-ID"]) %>/html" class="w-full mt-1 -mb-4 h-full" frameborder="0"></iframe>
               </div>
             </div>
           <% end %>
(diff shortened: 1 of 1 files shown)


lib/plug/templates/mailbox_viewer/index.html.eex
```diff
@@ -173,6 +173,32 @@
...
+      const iframe = document.querySelector("#html-mail");
+      if (iframe) {
+        Array.from(iframe.contentWindow.document.querySelectorAll("a")).forEach((link) => {
+          if (!link.hasAttribute("target")) {
+            link.setAttribute("target", "_top");
+          }
+        });
+      }
+
+      // make links in text only view clickable
```
chrismccord, line 187: Since we're in this space, can we also rewrite localhost links to match window location? This would be nice when serving behind a proxy in dev, ie ngrok or local https proxy:

```suggestion
      // make links in text only view clickable
      const textMail = document.querySelector("#text-body-content");
      if (textMail) {
        const textContent = textMail.textContent;
        // match common URL formats starting with http://, https://, or www.
        const urlRegex = /(https?:\/\/[^\s]+)|(www\.[^\s]+)/g;

        const htmlWithLinks = textContent.replace(urlRegex, function(url) {
          let href = url.startsWith('www.') ? 'https://' + url : url;

          // If link starts with localhost, rewrite it to match the current host/scheme/port
          try {
            const parsed = new URL(href);
            if (parsed.hostname === 'localhost') {
              parsed.protocol = window.location.protocol;
              parsed.hostname = window.location.hostname;
              parsed.port = window.location.port;
              href = parsed.toString();
            }
          } catch (e) {
            // Ignore errors for invalid URLs
          }

          return `<a href="${href}" style="text-decoration: underline;">${url}</a>`;
        });

        // Set the HTML content with clickable links
        textMail.innerHTML = htmlWithLinks;
      }
```

SteffenDE: @chrismccord I guess we could do that. What about having a default runtime.exs where you can set PHX_HOST=... PHX_SCHEME=https and then all URLs are right in the first place?

chrismccord: Still fails the ngrok case and my own more complex dynamic reverse proxy stuff

SteffenDE: I don't really care, so I'll let a swoosh maintainer decide

princemaple:
Apologies will get to this in a few days

princemaple:
Apologies will get to this in a few days

princemaple:
Back in Melbourne. Gonna merge this first and come back for improvements.

PR #1045, opened 2025-06-03
Fix unused variable warnings in CI - Gmail Test

DuldR:
I noticed there was an unused variable in the gmail test that was raising warnings in CI. 

image

This should make them go away.


princemaple:
Thanks!

PR #1049, opened 2025-06-18
Add deliver_many support to Brevo adapter

linusdm:
Implements batch email sending using Brevo's messageVersions API parameter. This allows sending multiple customized emails in a single API call.

Global-only parameters (shared across emails):
- sender
- attachment
- tags
- scheduledAt

Per-message parameters (customizable per email):
- to, cc, bcc
- replyTo
- subject, htmlContent, textContent
- templateId, params
- headers

Implementation detail: subject, htmlContent, textContent and templateId can be specified per message, but are required globally before overriding.


princemaple:
Thanks! This looks great.

PR #1058, opened 2025-07-26
Allow Customer.io to use TEMPLATE for from

axelson:
Patterned off of Sendingblue: https://github.com/swoosh/swoosh/issues/537

Also other minor tweaks:
- Add customer.io to the README (is there a defined order? I put it in a rough alphabetical order)
- Fix an errant `"`

PR #1059, opened 2025-07-29
Support rendering TEMPLATE in the mailbox

axelson:
Followup to https://github.com/swoosh/swoosh/pull/1058 that ensures that a `from: "TEMPLATE"` email doesn't break the mailbox preview. I'm guessing that anyone using `"TEMPLATE"` from https://github.com/swoosh/swoosh/issues/537 would also have a broken mailbox preview.

Question: Should we be more permissive here? Instead of supporting only `"TEMPLATE"`, instead we could allow any string


princemaple:
1.19.5

axelson:
Ah yeah that makes sense. Thank you!

PR #1070, opened 2025-09-25
Encode international domain names (IDNA)

jiegillet:
Hello! Sorry once again for an unsolicited PR.

Thank you for working on Swoosh, we just started using it at work, and we ran into a corner case: we have a number of German users who have emails with non-ASCII domain names, such as `_redacted_@hörexperte.de`.

This use case is usually covered by [IDNA](https://en.wikipedia.org/wiki/Internationalized_domain_name) and [RFC 3492](https://datatracker.ietf.org/doc/html/rfc3492), so I thought Swoosh could also handle it. Our previous email setup used the Erlang [`:idna`](https://hexdocs.pm/idna/) package to handle the encoding into Punycode, and since it was already a transitive dependency of hackney, I thought it was a good choice.

The encoding leaves ASCII domains untouched.

Let me know if I can make any changes to improve the PR.

### DIFF
--- a/lib/swoosh/email/render.ex
+++ b/lib/swoosh/email/render.ex
@@ -2,8 +2,8 @@ defmodule Swoosh.Email.Render do
   @moduledoc false
 
   def render_recipient(nil), do: ""
-  def render_recipient({nil, address}), do: address
-  def render_recipient({"", address}), do: address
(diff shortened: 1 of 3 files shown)


princemaple:
I will go through it again when my brain is fresh. Thanks for the PR!

Issue #1076, opened 2025-10-25
Add support for reply_to in Scaleway adapter


jaimeiniesta:
The Scaleway adapter ignores `reply_to`

https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/scaleway.ex#L108

Would it be possible to add support to it?

jaimeiniesta:
Hi again, I found this, maybe we can do it like that?

https://github.com/scaleway/scaleway-sdk-js/issues/848

jaimeiniesta:
Sure, I'll work on this.

jaimeiniesta:
On top of this we could also add support for the `reply_to` abstraction, it would set the appropriate "Reply-To" header.

PR #1077, opened 2025-10-26
add support for additional_headers provider option in Scaleway

jaimeiniesta:
Related to https://github.com/swoosh/swoosh/issues/1076


jaimeiniesta:
Sorry, I'm not sure what you mean by using the headers clause directly, can you explain a bit more?

jaimeiniesta:
Alright, once you've checked this part, I'll work on adding support for the `reply_to` clause.

princemaple:
This is why I got confused. Scaleway adapter has 2 prepare_headers.

prepare_headers/1: https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/scaleway.ex#L70-L77
prepare_headers/2: https://github.com/swoosh/swoosh/blob/main/lib/swoosh/adapters/scaleway.ex#L145-L149

We should rename the first one to `request_headers`.

And `"headers"` key in payload does not seem valid, according to the docs. I think we can use this clause directly, changing it to `"additional_headers"`.

PR #1080, opened 2025-11-26
Add Resend adapter to README

jtormey:
I maintain the Resend library for Elixir, and it includes a Swoosh adapter. I noticed that the adapters list only has adapters implemented by Swoosh, but was curious if 3rd-party adapters are welcome in the documentation. Thanks!


jtormey:
Updated!

PR #1089, opened 2025-12-24
feat: add a resend adapter

ceolinrenato:
I know there's a community Resend adapter https://github.com/elixir-saas/resend-elixir

But I realized it's not just a Swoosh adapter, it has plenty more stuff. For a swoosh adapter it looks like we only need implementations of deliver/2 and deliver_many/2

Which are https://resend.com/docs/api-reference/emails/send-email and https://resend.com/docs/api-reference/emails/send-batch-emails respectively

So I figured we might as well include a built in adapter for it that uses the `Swoosh.ApiClient` module like the other adapters we have

### DIFF
--- a/lib/swoosh/adapters/resend.ex
+++ b/lib/swoosh/adapters/resend.ex
@@ -109,7 +109,6 @@ defmodule Swoosh.Adapters.Resend do
 
   defp base_url(config), do: config[:base_url] || @base_url
 
-  @impl Swoosh.Adapter
   def deliver(%Email{} = email, config \\ []) do
     headers = prepare_request_headers(config, email)
     body = email |> prepare_body() |> Swoosh.json_library().encode!()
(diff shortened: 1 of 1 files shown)


princemaple:
I'm not entirely sure if I want to maintain something that already has an official client. 🤔

What would you do @josevalim ?

ceolinrenato:
> > Good points. Could you also update README to list this adapter. Add link to batch api in module docs. And, could I ping you on issues / PRs about this adapter? ;)
> 
> For sure!
> 
> on it
> 
> Also I realized there's some commits removing `@impl` because of docs like: [9224814](https://github.com/swoosh/swoosh/commit/922481405441c4f9066432dd321ea04eec569b6f) should I do the same thing or is that not a thing on current elixir versions anymore?

yeah I just realized it doesn't get listed by running `mix doc` locally, removed it on https://github.com/swoosh/swoosh/pull/1089/commits/cf31df4f070e39d802470767ba0f611c84736d9d

princemaple:
Thanks. Beat me to it.

PR #1096, opened 2026-01-16
Fix Content-Transfer-Encoding for message/* attachments

romsahel:
## Problem

The SMTP adapter currently hardcodes `Content-Transfer-Encoding: base64` for all attachments in `prepare_attachment/1` (lib/swoosh/adapters/smtp/helpers.ex:212, 224).

This is incorrect for `message/rfc822` content types. Per [RFC 2046 Section 5.2.1](https://datatracker.ietf.org/doc/html/rfc2046#section-5.2.1):

> No encoding other than "7bit", "8bit", or "binary" is permitted for the body of a "message/rfc822" entity.

As well as for other `message/*` content types:
> entities of type "message/partial" must always have a content-transfer-encoding of 7bit (the default).

> MIME entities of type "message/external-body" MUST have a content-transfer-encoding of 7bit (the default).

When forwarding emails as attachments (a common use case), the current implementation produces invalid MIME messages that may be rejected or incorrectly parsed by receiving mail servers (although it is actually widely supported, some clients refuse to parse them correctly).

## Solutions Considered

1. **Check the `content_type`**: If content_type starts with "message/", use "7bit" instead of "base64"
   - Pro: No public API change, automatic for message/* types
   - Con: Not configurable for other types of attachments

2. **Allow encoding override via custom headers**: Reorder header handling so custom headers can override the default
   - Pro: No public API change
   - Con: Flaky and hidden configurable behavior

3. **Add a `transfer_encoding` field to `Swoosh.Attachment`**: Let users explicitly specify the encoding
   - Pro: Configurable for all types of attachments
   - Con: Change to public API, more impactful change, not automatic for message/rfc822 attachments

## Implemented Solution

We chose solution 1 as it is the least invasive change with no modifications to the public `Swoosh.Attachment` API. 
Note that in our tests, the problem does not manifest itself when using the Mailgun adapter since the Mailgun API fixes the message on their side.

Added `transfer_encoding_for_content_type/1` in `lib/swoosh/adapters/smtp/helpers.ex`:
- `message/rfc822` → `"8bit"` (allows 7bit, 8bit, or binary per Section 5.2.1)
- `message/*` (other) → `"7bit"` (required for message/partial and message/external-body per Sections 5.2.2, 5.2.3)
- everything else → `"base64"`

### DIFF
--- a/lib/swoosh/adapters/smtp/helpers.ex
+++ b/lib/swoosh/adapters/smtp/helpers.ex
@@ -202,6 +202,7 @@ defmodule Swoosh.Adapters.SMTP.Helpers do
        ) do
     [type, format] = String.split(content_type, "/")
     content = Swoosh.Attachment.get_content(attachment)
+    transfer_encoding = transfer_encoding_for_content_type(content_type)
 
     case attachment_type do
       :attachment ->
@@ -209,7 +210,7 @@ defmodule Swoosh.Adapters.SMTP.Helpers do
           type,
           format,
           [
-            {"Content-Transfer-Encoding", "base64"}
+            {"Content-Transfer-Encoding", transfer_encoding}
             | custom_headers
           ],
           attachment_content_params(:attachment, filename),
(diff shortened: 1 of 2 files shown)


princemaple:
Thanks for the detailed description!

romsahel:
Perfect! Thank you for the quick response 🙏

Issue #1131, opened 2026-05-08
Bump hackney dep


allenwyma:
Hackney is already on version 4.0, while hackney here is still on version 1.

Now that `ex_aws` was already updated to use 3 or 4 for Hackney, can we consider to support hackney 4?

princemaple:
Hello, I haven't been paying attention to where Hackney has been going. If the code change is not drastic, PRs are more than welcome to add support for higher versions of Hackney. Cheers!

allenwyma:
Okay, I'll see if i can get some time. If i update the codebase, will the automated tests catch any issues?

princemaple:
Probably. We only use post anyway.

allenwyma:
@princemaple done!

seems post didn't change:

https://hexdocs.pm/hackney/hackney.html#post/1
https://hexdocs.pm/hackney/1.25.0/hackney.html#post/1

princemaple:
Is there any version that should NOT be used? There has to be a reason it bumped so many times in such short period.

allenwyma:
@princemaple looks like 4.0 removed metrics and 3.0 changes a bit of the response:

```erlang
%% Before (2.x) - HTTP/1.1
{ok, 200, Headers, ConnPid} = hackney:get(URL),
{ok, Body} = hackney:body(ConnPid).

%% After (3.x) - All protocols
{ok, 200, Headers, Body} = hackney:get(URL).
```

and 2.0 added new functions

Maybe we'll need to check the post response. I ran mix test and all ran fine.

allenwyma:
Although your post seems to match the new style:

`{:ok, status, headers, body} | {:error, term()}`

princemaple:
Cheers. I'll do a deeper dive before merge. Thanks

allenwyma:
@princemaple let me know if i can help. 😄

PR #1132, opened 2026-05-08
update to support hackney less than 5.0

allenwyma:
relates to #1131


grzuy:
For what is worth, note that while this did expanded `hackney` compatibility range in `mix.exs` for swoosh, it didn't actually updated the used hackney version in `mix.lock`, which means the test cases in this PR and any development on swoosh is still running against hackney 1.x.

allenwyma:
> For what is worth, note that while this did expanded `hackney` compatibility range in `mix.exs` for swoosh, it didn't actually updated the used hackney version in `mix.lock`, which means the test cases in this PR and any development on swoosh is still running against hackney 1.x.

Good point. Can have a look and run again locally

allenwyma:
> For what is worth, note that while this did expanded `hackney` compatibility range in `mix.exs` for swoosh, it didn't actually updated the used hackney version in `mix.lock`, which means the test cases in this PR and any development on swoosh is still running against hackney 1.x.

merged with upstream, updated ex_aws to latest, hackney was upgraded to 4.0.1, mix test passed all tests :)

PR #1152, opened 2026-05-28
Added adapter for self-hosted Mailpit

waseigo:
Added a new adapter for self-hosted [Mailpit](https://mailpit.axllent.org/) servers.

- Optional support for Basic Auth
- `mailpit_test.exs` passes
- Also tested from within IEx with my own self-hosted Mailpit v1.30.1 server with Basic Auth


princemaple:
Thanks!

princemaple:
Hey would like to take a look at this #1154

PR #1155, opened 2026-06-03
fix fat-fingering `content_id` instead of `cid`, added tests, fixed outdated expected response in `mua_test.exs`

waseigo:
- I had typed `attachment.content_id` instead of `attachment.cid`; fixed.
- I added 2 new tests for `deliver/2` a) with an inline attachment and b) with an inline attachment and custom `cid`.
- I ran `mix test --only mailpit` (hadn't noticed that this exists since last year for testing with a local docker container). The `"Text"` field with the latest docker image doesn't include a trailing `\r\n` so I adapted the expected response.

Verification:
```
$ docker run -d --rm -p 1025:1025 -p 8025:8025 --name mailpit axllent/mailpit
$ mix test --only mailpit
Running ExUnit with seed: 620926, max_cases: 24
Excluding tags: [:test, :integration]
Including tags: [:mailpit]
...
Finished in 1.2 seconds (1.2s async, 0.00s sync)
29 doctests, 594 tests, 0 failures, 617 excluded
$ mix test 
Running ExUnit with seed: 764694, max_cases: 24
Excluding tags: [:integration, :mailpit]
...
Finished in 1.9 seconds (1.8s async, 0.06s sync)
29 doctests, 594 tests, 0 failures, 22 excluded
```

Should be OK now.

### DIFF
--- a/lib/swoosh/adapters/mailpit.ex
+++ b/lib/swoosh/adapters/mailpit.ex
@@ -173,7 +173,7 @@ defmodule Swoosh.Adapters.Mailpit do
 
         case attachment.type do
           :inline ->
-            content_id = attachment.content_id || attachment.filename
+            content_id = attachment.cid || attachment.filename
             Map.put(att_map, "ContentID", content_id)
 
           _ ->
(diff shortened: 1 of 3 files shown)


princemaple:
Cheers!

PR #1166, opened 2026-06-25
Fix Postmark adapter not adding the Recipient's name to the Reply-To header

vickz84259:
As the title states, the Postmark adapter did not add the Recipient's name when calling `reply_to/2`. 

Before, calling `reply_to({"Victor", "victor@example.com"})` would result in the header: `Reply-To: victor@example.com`.

With this fix we now get: `Reply-To: Victor <victor@example.com>`


princemaple:
Thanks!

PR #1177, opened 2026-07-25
Add AhaSend adapter

Sameer1122:
Adds an adapter for [AhaSend](https://ahasend.com), sending through their HTTP API (`POST /v2/accounts/{account_id}/messages`).

### Features

- `deliver/2` and `deliver_many/2`
- Attachments, including inline (`content_id` + `inline` disposition)
- Provider options: `tags`, `substitutions`, `tracking`, `retention`, `schedule`, `sandbox`, `sandbox_result`, `idempotency_key`
- Typed error tuples for non-2xx responses

### `deliver_many/2`

Sends each email as its own request so one invalid recipient can't affect the others — AhaSend validates the whole recipients array up front and rejects the entire request on a single bad address. Results are returned in input order; delivery halts on the first failure.

### Notes

- Config requires `:api_key` and `:account_id` (the account is in the URL path).
- Emails with `cc`/`bcc` use the conversation endpoint; others use the standard send endpoint.
- Verified against the live AhaSend sandbox: `deliver`, attachments, cc/conversation, `deliver_many`, and the error paths.

25 tests, `mix format` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### ✨ Features
+
+- Add AhaSend adapter @Sameer1122 (#1177)
+
 ## 1.26.3
 
 ### 🔒 Security
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ included:
 | Lettermint                   | [Swoosh.Adapters.Lettermint](https://hexdocs.pm/swoosh/Swoosh.Adapters.Lettermint.html#content)                                 |                  |
 | Resend                       | [Swoosh.Adapters.Resend](https://hexdocs.pm/swoosh/Swoosh.Adapters.Resend.html#content)                                         |                  |
 | Azure Communication Services | [Swoosh.Adapters.AzureCommunicationServices](https://hexdocs.pm/swoosh/Swoosh.Adapters.AzureCommunicationServices.html#content) |                  |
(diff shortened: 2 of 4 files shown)


princemaple:
Hi @Sameer1122 Thanks for the PR. How popular is this service? Do you think it's worth going into Swoosh instead of staying as a separate package?

Sameer1122:
@Princemaple Thanks for taking a look! On popularity: AhaSend is a European transactional email provider based in Amsterdam, running on our own infrastructure across several EU countries. We're growing steadily and see real demand from the Elixir side specifically, which is why we built this: EU data residency and GDPR are a recurring reason European teams move off US providers, and Swoosh is one of the places where they look first. Being in-tree obviously makes us discoverable for people who'd otherwise not know we exist, which is the main reason I'd prefer it, but I understand if you'd rather keep the surface area small. The API surface is small and stable (a single send endpoint, standard auth), so the maintenance burden should stay low.

princemaple:
How would you compare AhaSend to Lettermint?

MarkNLD:
@princemaple Hi Po, I am Mark, one of AhaSend's founders. AhaSend and Lettermint both operate in the same space. Just like Resend, Postmark and Sendgrid do in the USA. Currently AhaSend and Lettermint both have real traction in the European Union because of companies switching to European solutions.

We are adding integrations like the Swoosh Adapters to minimise the impact of a migration for these companies.

Thank you for your time! And if you'd like to have a chat, feel free to reach out in our Discord (https://discord.com/invite/WN4puKFqAA).

PR #1186, opened 2026-08-13
Return empty text for a missing XML node instead of crashing (#1185)

### LINKED ISSUE #1185: Swoosh.Adapters.AmazonSES crashes with FunctionClauseError (xmerl_xpath.node_type/1)
### Have you tried the latest version of Swoosh?

- [x] Yes

### Describe the bug

We have an occurrence where `Swoosh.Adapters.AmazonSES.parse_error_response` crashes when it is not able to correctly parse the error response from SES (or an intermediary).

I suspect that this could be invalid or non-XML returned.

Here is a stack trace:
```
FunctionClauseError
no function clause matching in :xmerl_xpath.node_type/1
mechanism: generic
handled: true

Stacktrace:
  xmerl_xpath.erl:647 in :xmerl_xpath.node_type/1
  xmerl_xpath.erl:149 in :xmerl_xpath.string/5
  lib/swoosh/adapters/xml/helpers.ex:38 in Swoosh.Adapters.XML.Helpers.text/1
  lib/swoosh/adapters/amazon_ses.ex:147 in Swoosh.Adapters.AmazonSES.parse_error_response/1
  lib/swoosh/adapters/amazon_ses.ex:128 in Swoosh.Adapters.AmazonSES.deliver/2
  lib/my_app/mailer.ex:5 in anonymous fn/2 in AppName.Mailer.instrument/3
  deps/telemetry/src/telemetry.erl:359 in :telemetry.span/3
  lib/my_app/mailer/user_notifier.ex:49 in anonymous fn/4 in MyApp.Mailer.deliver_email/2
```

When SES returns with a non-XML response on an error code.

### Expected behavior

Not sure what the expected behavior should be, perhaps returning an `{:error, :unknown_error}` code in the error parser? Or Logging the error as a first step. Please advice and I would be happy to take look 🙏 

### Your reproducible example

_No response_

### Screenshots or Videos

_No response_

### Platform

- OTP: 29
- Elixir: 1.20.3

### Additional context

_No response_

youdie006:
Fixes #1185.

### Problem

`Swoosh.Adapters.XML.Helpers.first/2` returns `nil` when its xpath matches no node, and `text/1` then passed that `nil` straight to `:xmerl_xpath.string/2`, raising:

```
** (FunctionClauseError) no function clause matching in :xmerl_xpath.node_type/1
    (xmerl) xmerl_xpath.erl: :xmerl_xpath.node_type/1
    (swoosh) lib/swoosh/adapters/xml/helpers.ex:38: Swoosh.Adapters.XML.Helpers.text/1
```

This crashes `AmazonSES.parse_error_response/1`, which runs on any HTTP status `> 399` and calls `first_text(node, "//Error/Code")` / `first_text(node, "//Message")`. Whenever the SES (or an intermediary) error body is valid XML that lacks those elements, delivering an email raises `FunctionClauseError` instead of returning a clean `{:error, ...}`.

### Fix

Add a `text(nil)` clause returning `""` — a missing node has no text. `parse_error_response/1` then returns `%{code: ..., message: ""}` instead of crashing.

### Test

Added a regression test in `test/swoosh/adapters/xml/helpers_test.exs`: `first_text` on a path that matches no node returns `""`. Red-green verified with `mix test`: before the fix the test raises the exact `FunctionClauseError` in `:xmerl_xpath.node_type/1` from the issue; after, it returns `""`. `mix format --check-formatted` is clean.

---
Disclosure: I used AI assistance (Claude) while preparing this change. I verified the crash, ran the tests (red-green), and take responsibility for the contribution.

### DIFF
--- a/lib/swoosh/adapters/xml/helpers.ex
+++ b/lib/swoosh/adapters/xml/helpers.ex
@@ -32,6 +32,10 @@ defmodule Swoosh.Adapters.XML.Helpers do
   end
 
   @doc false
+  # A missing node (no xpath match, so `first/2` returned nil) has no text.
+  # Return an empty string instead of crashing in :xmerl_xpath.node_type/1.
+  def text(nil), do: ""
+
   def text(node) do
     "./text()"
     |> to_charlist
--- a/test/swoosh/adapters/xml/helpers_test.exs
+++ b/test/swoosh/adapters/xml/helpers_test.exs
@@ -54,4 +54,13 @@ defmodule Swoosh.Adapters.XML.HelpersTest do
 
     assert text == ""
   end
+
+  test "first_text returns blank when the path matches no node", %{xml_string: xml_string} do
+    text =
+      xml_string
+      |> XMLHelper.parse()
+      |> XMLHelper.first_text("//missing")
+
+    assert text == ""
+  end
 end


princemaple:
Thanks!

PR #1190, opened 2026-08-22
Add TurboSMTP adapter

NewtTheWolf:
Adds an adapter for the [TurboSMTP](https://serversmtp.com) API.

I am contributing this on behalf of TurboSMTP, so the provider-specific behaviour below is verified against the live API rather than read off the docs.

```elixir
  adapter: Swoosh.Adapters.TurboSMTP,
  consumer_key: "my-consumer-key",
  consumer_secret: "my-consumer-secret"
```

`POST /api/v2/mail/send`, credentials as `consumerKey` / `consumerSecret` headers. TurboSMTP has exactly two regions; the EU one is reached by pointing `base_url` at `https://api.eu.turbo-smtp.com/api/v2`, which follows what Brevo, Scaleway and Lettermint already do rather than adding a new config key.

## Two provider quirks the adapter absorbs

**Inline images.** TurboSMTP composes each inline part's Content-ID as `<content_id@sender-domain>` while `content_id` travels bare, so Swoosh's documented `` never matches and the image is delivered as an ordinary attachment. The send returns 200 either way, which is what makes it hard to spot. The adapter qualifies `cid:` references with the sender's domain and leaves a `cid` that already carries one alone.

Verified end to end against the live API with two otherwise identical mails. Only the qualification differs:

| | HTML reference | Thunderbird reports |
|---|---|---|
| through the adapter | `cid:swoosh.png@sender-domain` | 1 embedded image, rendered in the body |
| bare `cid:` | `cid:swoosh.png` | 1 file attached, body gap |

**Commas in recipient display names.** `to`, `cc` and `bcc` are split on commas before RFC 5322 quoted strings are parsed, so `{"Doe, John", "j@example.com"}` is torn apart. The live API answers:

```
400 {"message":"error","errors":[["'\"Doe' 'to' email not valid"]]}
```

That names a fragment of the display name, so the adapter returns `{:error, {:invalid_recipient, name}}` before sending instead. `from` and `reply_to` are not comma-split and keep accepting it.

## Verification

`mix test`, `mix format --check-formatted` and `mix deps.unlock --check-unused` are green on Elixir 1.20, and the suite also passes on 1.18. 14 adapter tests cover the payload mapping, both quirks, provider options, and the 4xx/5xx paths.

Live sends against `api.turbo-smtp.com`, all returning real message ids: plain text, HTML with Reply-To and a custom header, an inline image, and one through the EU host.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### ✨ Features
+
+- Add TurboSMTP adapter @NewtTheWolf
+
 ## 1.27.1
 
 ### 🐛 Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### ✨ Features
 
-- Add TurboSMTP adapter @NewtTheWolf
+- Add TurboSMTP adapter @NewtTheWolf (#1190)
 
 ## 1.27.1
 
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ included:
 | Resend                       | [Swoosh.Adapters.Resend](https://hexdocs.pm/swoosh/Swoosh.Adapters.Resend.html#content)                                         |                  |
 | Azure Communication Services | [Swoosh.Adapters.AzureCommunicationServices](https://hexdocs.pm/swoosh/Swoosh.Adapters.AzureCommunicationServices.html#content) |                  |
 | AhaSend                      | [Swoosh.Adapters.AhaSend](https://hexdocs.pm/swoosh/Swoosh.Adapters.AhaSend.html#content)                                       | EU               |
+| TurboSMTP                    | [Swoosh.Adapters.TurboSMTP](https://hexdocs.pm/swoosh/Swoosh.Adapters.TurboSMTP.html#content)                                   | EU               |
 | ------                       | **Below are not fully featured services**                                                                                       | ------           |
 | Loops                        | [Swoosh.Adapters.Loops](https://hexdocs.pm/swoosh/Swoosh.Adapters.Loops.html#content)                                           |                  |
 | PostUp                       | [Swoosh.Adapters.PostUp](https://hexdocs.pm/swoosh/Swoosh.Adapters.PostUp.html#content)                                         |                  |
--- a/lib/swoosh/adapters/turbo_smtp.ex
+++ b/lib/swoosh/adapters/turbo_smtp.ex
@@ -95,10 +95,10 @@ defmodule Swoosh.Adapters.TurboSMTP do
       url = [base_url(config), @api_endpoint]
 
       case Swoosh.ApiClient.post(url, headers, body, email) do
-        {:ok, 200, _headers, body} ->
-          handle_success(body)
+        {:ok, code, _headers, body} when code in 200..299 ->
+          handle_success(code, body)
 
-        {:ok, code, _headers, body} when code >= 400 ->
+        {:ok, code, _headers, body} ->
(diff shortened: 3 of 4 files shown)


princemaple:
Pushed follow-up commits (7bc7875, 6b444b6) addressing part of the review feedback:

- **Multiple `reply_to` addresses**: `custom_headers/1` previously crashed with `FunctionClauseError` if `reply_to` was a list (e.g. `reply_to([{"Pepper Potts", "pepper@example.com"}, "happy@example.com"])`). It now joins formatted addresses into a single comma-separated `Reply-To` header.
- **Inline cid qualification efficiency**: `qualify/3` compiled a fresh regex and rescanned the whole HTML body once per inline attachment. It now builds a single alternation pattern and does one compile + one `Regex.replace` pass for all cids.
- **Ordering dependency documented**: added a comment in `prepare_payload/1` noting that `qualify_inline_cids` must run after `prepare_attachments` (it reads `payload["attachments"]`), so a future reordering doesn't silently break inline images.

Added two tests (multiple reply_to addresses, multiple inline attachments qualified in one pass). Full suite passes (646 tests).

(An earlier version of this commit also extracted a shared `decode_json/1` helper for `handle_success/1`/`decode_body/1` — reverted that, not worth the indirection for two call sites.)

Two other findings from review were left as-is for now and can be follow-ups:
- `deliver/2`'s case statement only matches status `200` exactly or `>= 400`, so a 2xx-non-200 or 3xx response would raise `CaseClauseError` instead of returning a result tuple.
- `prepare_from/2` crashes with `FunctionClauseError` if `email.from` is left `nil`.

princemaple:
Hi @NewtTheWolf , we (claude and I) made some changes. If you are happy with them, I think this PR is ready to go in.

(I also checked that TurboSMTP has been around for a long time, so it's OK to include it)

NewtTheWolf:
Thanks for the follow-ups, all three are good and I have pulled them in.

The `reply_to` list clause was a genuine bug on my side. `Swoosh.Email` types the field as `[mailbox] | mailbox | nil` and `reply_to/2` has its own `is_list` clause, so my `custom_headers/1` would have raised for anything a caller is entitled to pass.

I checked the single-alternation rewrite against the case that can go wrong, one cid being a prefix of another. For `cid:ab` with the pattern `a|ab`, the engine matches `a`, the `(?![\w.@-])` lookahead fails on `b`, it backtracks and takes `ab`. Same result as the sequential version, one pass instead of one per attachment.

On the two you left open:

**Unexpected statuses** — fixed in ccabe4c, and it turned out to be two problems rather than one. Success now matches `code in 200..299` with a catch-all error clause below it, so `deliver/2` can no longer raise instead of returning a result tuple. While changing it I noticed `handle_success/1` also hardcoded `200` in the error tuple when the body would not decode, so an undecodable 2xx reported the wrong status; it carries the real code now.

Two tests cover it. I checked they actually bite by reverting the clause and re-running: both fail with the `CaseClauseError` they exist to prevent.

**`from` being `nil`** — I would leave this one alone. `Swoosh.Mailer.deliver/2` already answers `{:error, :from_not_set}` in `mailer.ex:196` before any adapter runs, and no adapter in the repo duplicates that guard. It is only reachable by calling `TurboSMTP.deliver/2` directly and bypassing the mailer, so a guard here would add a second place to maintain the same rule.

Full suite green (648), format and `deps.unlock --check-unused` clean on 1.20.

0x8fv:
@NewtTheWolf Kp was .ex ist aber sieht richtig aus. Ist zwar schon merged aber hiermit habt ihr auch meine Genehmigung um es zu mergen. Das nächste mal bitte mit mir abklären! 😁👍

PR #1196, opened 2026-09-05
Add MailKite adapter

bucabay:
Adds an adapter for the [MailKite](https://mailkite.dev) API.

I work on MailKite, so this is a vendor contribution and I'll maintain the adapter going forward (API changes, bug reports, provider options).

```elixir
  adapter: Swoosh.Adapters.MailKite,
  api_key: "mk_live_..."
```

`POST https://api.mailkite.dev/v1/send`, Bearer key, JSON body. Same shape as the Lettermint / AhaSend adapters: `from`/`to`/`cc`/`bcc`/`reply_to`/`subject`/`text_body`/`html_body`/`headers`/attachments map onto the documented request fields, `base_url` is overridable for tests, and the response is normalised to `{:ok, %{id: "msg_…", status: "sent"}}`. The API accepts RFC 5322 mailboxes, so `{name, address}` tuples are rendered as `Name <address>` (quoted when the name is not a plain atom run — same rule as our nodemailer transport).

## Two API details the adapter absorbs

**Reply-To is a single field.** MailKite's `replyTo` is one string, while `Swoosh.Email.reply_to/2` also accepts a list. A list is joined with `", "` — the way a mail client would render the header — instead of dropping addresses or raising. Covered by a test.

**No inline (CID) attachments.** MailKite attachments carry no `Content-ID`, so an email with a `type: :inline` attachment cannot be expressed. Rather than silently sending it as a regular attachment (which would leave a broken `cid:` reference in the HTML), `deliver/2` returns `{:error, :inline_attachments_not_supported}` before any request is made, and the moduledoc says so. Regular attachments are sent base64-inline.

## Provider options

`metadata`, `template_id` / `template_data`, `in_reply_to`, `scheduled_at`, `track_opens`, `track_clicks` — each maps to the corresponding camelCase API field, documented in the moduledoc.

## Testing

- `test/swoosh/adapters/mailkite_test.exs`: 18 tests against Bypass (request shape for text/html/both, names + quoting, cc/bcc, multi reply-to, custom headers, attachment, inline-attachment error, provider options, template send, scheduled send, 400/401/5xx, `validate_config/1`).
- `mix test`: 666 passed, 0 failures on this branch (Elixir 1.20 / OTP 28).
- `mix compile --warnings-as-errors` and `mix format --check-formatted` clean.

README adapter table and CHANGELOG (Unreleased) updated. No changes outside the new adapter and its test.

Happy to adjust anything to match house style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### ✨ Features
+
+- Add MailKite adapter @bucabay
+
 ## 1.28.0
 
 ### ✨ Features
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ included:
 | Azure Communication Services | [Swoosh.Adapters.AzureCommunicationServices](https://hexdocs.pm/swoosh/Swoosh.Adapters.AzureCommunicationServices.html#content) |                  |
 | AhaSend                      | [Swoosh.Adapters.AhaSend](https://hexdocs.pm/swoosh/Swoosh.Adapters.AhaSend.html#content)                                       | EU               |
 | TurboSMTP                    | [Swoosh.Adapters.TurboSMTP](https://hexdocs.pm/swoosh/Swoosh.Adapters.TurboSMTP.html#content)                                   | EU               |
+| MailKite                     | [Swoosh.Adapters.MailKite](https://hexdocs.pm/swoosh/Swoosh.Adapters.MailKite.html#content)                                     |                  |
 | ------                       | **Below are not fully featured services**                                                                                       | ------           |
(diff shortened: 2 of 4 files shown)


princemaple:
Hi @bucabay thanks for the PR. I'm afraid this service is too new for me to consider the merge.

bucabay:
Can I get back to you in 6 months with the PR?

princemaple:
Sure thing. We can definitely reevaluate in 6 months.

bucabay:
Tracking note: this PR is paused pending a 6-month recheck, as discussed with @princemaple above.

- Target recheck date: **March 21, 2027** (closed Sep 21, 2026 + 6 months)
- Plan: reopen/resubmit the MailKite adapter if the API and service are still live, with updated docs/tests if anything changed

PR #1202, opened 2026-09-17
Raise a clear error for SMTP builds compiled without :gen_smtp

### LINKED ISSUE #1189: SMTP adapter compiled without the optional :gen_smtp silently produces a build that raises BadMapError on every send
### Have you tried the latest version of Swoosh?

- [X] Yes

### Describe the bug

`Swoosh.Adapters.SMTP.Helpers` decides the **shape** of the mimetuple's parameters element at
**compile time**, from whether `:gen_smtp_client` happened to be loadable at that moment
([`helpers.ex:21-38`](https://github.com/swoosh/swoosh/blob/ffa34f6b958d747e31fc0ce53328c8842ee75d4b/lib/swoosh/adapters/smtp/helpers.ex#L21-L38)):

```elixir
# TODO: Remove conditional handling when going 2.0
gen_smtp_major =
  if Code.ensure_loaded?(:gen_smtp_client) do
    Application.load(:gen_smtp)
    :gen_smtp |> Application.spec(:vsn) |> to_string() |> Version.parse!() |> Map.get(:major)
  else
    0
  end

@parameters if(gen_smtp_major >= 1, do: %{}, else: [])

defp mime_encode(type, subtype, headers, parts, encoding_config) do
  :mimemail.encode({type, subtype, headers, @parameters, parts}, encoding_config)
end
```

`:gen_smtp` is [`optional: true`](https://github.com/swoosh/swoosh/blob/ffa34f6b958d747e31fc0ce53328c8842ee75d4b/mix.exs#L78),
so the SMTP adapter is compiled for **every** Swoosh user, gen_smtp present or not. Compiled while
gen_smtp is absent, `@parameters` freezes to `[]` — and so do its per-part twin
[`@content_params`](https://github.com/swoosh/swoosh/blob/ffa34f6b958d747e31fc0ce53328c8842ee75d4b/lib/swoosh/adapters/smtp/helpers.ex#L153-L164)
and both
[`attachment_content_params/2` clause pairs](https://github.com/swoosh/swoosh/blob/ffa34f6b958d747e31fc0ce53328c8842ee75d4b/lib/swoosh/adapters/smtp/helpers.ex#L243-L275).

gen_smtp 1.x reads that slot as a **map** —
[`maps:get(content_type_params, Parameters, [])`](https://github.com/gen-smtp/gen_smtp/blob/1fa4ffc1e96fe48885990c6ac43f50edb7de2dff/src/mimemail.erl#L744)
and the `transfer_encoding` / `disposition` / `disposition_params` reads at
[`:812-830`](https://github.com/gen-smtp/gen_smtp/blob/1fa4ffc1e96fe48885990c6ac43f50edb7de2dff/src/mimemail.erl#L812-L830),
all inside
[`ensure_content_headers/7`](https://github.com/gen-smtp/gen_smtp/blob/1fa4ffc1e96fe48885990c6ac43f50edb7de2dff/src/mimemail.erl#L731).
So if gen_smtp is present **at runtime**, a `[]` in that slot means *every* send raises:

```
** (BadMapError) expected a map, got:

    []

    (gen_smtp 1.3.0) :mimemail.ensure_content_headers/7
    (gen_smtp 1.3.0) .../deps/gen_smtp/src/mimemail.erl:204: :mimemail.encode/2
    repro.exs:8: (file)
```

Exactly one of the four combinations is silent:

| gen_smtp at compile time | at runtime | result |
| --- | --- | --- |
| present | present | fine |
| absent | absent | fails loudly and correctly (undefined `:gen_smtp_client.send/2`) |
| present | absent | likewise loud |
| **absent** | **present** | **this bug** — every send raises `BadMapError` |

The mis-attribution is the real cost. The top stack frames are gen_smtp's, so it reads as a gen_smtp
bug or a malformed email; nothing points at the stale Swoosh artifact that actually caused it. And
inspecting the email teaches you nothing, because the offending value is a module attribute baked in
at compile time, not anything in the message.

Worth noting for the fix: the bad state is **trivially detectable at runtime**. On the poisoned build
below, `Code.ensure_loaded?(:gen_smtp_client)` is `true` and `Application.spec(:gen_smtp, :vsn)` is
`"1.3.0"` — while the compiled `@parameters` is `[]`.

Deterministic, from a fresh project. Verified on swoosh 1.27.0 + gen_smtp 1.3.0, Elixir 1.20.1 /
OTP 29, Linux:

```bash
mix new smtp_probe && cd smtp_probe
```

`mix.exs` — deps:

```elixir
{:swoosh, "~> 1.27"},
{:gen_smtp, "~> 1.3"},
```

`config/config.exs` (only to keep `Swoosh.Application` from wanting an API client):

```elixir
import Config
config :swoosh, api_client: false
```

`repro.exs`:

```elixir
email =
  Swoosh.Email.new()
  |> Swoosh.Email.from({"Sender", "sender@example.com"})
  |> Swoosh.Email.to({"Recipient", "recipient@example.com"})
  |> Swoosh.Email.subject("Hello")
  |> Swoosh.Email.text_body("Body")

Swoosh.Adapters.SMTP.Helpers.body(email, [])
```

Then:

```bash
mix deps.get && mix compile
mix run repro.exs            # OK — encodes fine

# reproduce a swoosh compiled while :gen_smtp was not loadable
mv _build/dev/lib/gen_smtp/ebin/gen_smtp_client.beam /tmp/
mix deps.compile swoosh --force
mv /tmp/gen_smtp_client.beam _build/dev/lib/gen_smtp/ebin/

mix run repro.exs            # ** (BadMapError) expected a map, got: []
```

Nothing about the message changed between the two `mix run`s — only which Swoosh artifact is in
`_build`. (`Swoosh.Adapters.SMTP.Helpers.body/2` is exactly what `Swoosh.Adapters.SMTP.deliver/2`
calls before it touches a socket, so this is the same failure a real send hits, minus the relay.)

**How it arrives in the wild, with nobody hiding a `.beam`:** an existing project that already has a
compiled Swoosh in `_build` adds `:gen_smtp` and runs `mix deps.get && mix compile`. Swoosh's own
lock entry did not change, and an **optional** dependency's appearance does not invalidate the
dependent, so Mix has no reason to recompile Swoosh — the artifact compiled against gen_smtp's
absence survives, and every send from then on raises. That is how we hit it: `:gen_smtp` entered the
lockfile, a warm CI `_build` kept the old Swoosh, and SMTP delivery broke with a stacktrace pointing
into `:mimemail`. `mix deps.compile swoosh --force` (per `MIX_ENV`) is the repair — once you know
that is what you are looking at.

### Expected behavior

That a build in this state either cannot happen, or says so.

**The ask: make the mismatch self-diagnosing.** Where `@parameters == []` but the *runtime*
`:gen_smtp` major is `>= 1`, raise an error naming the cause and the repair
(`mix deps.compile swoosh --force`, per `MIX_ENV`) instead of letting `:mimemail` fail with a
`BadMapError` several frames away. That is one comparison per send, zero behaviour change for
correct builds, and no policy decision to make. A compile-time `raise` is not available for the same
reason the bug exists: at compile time gen_smtp is *legitimately* absent for everyone not using the
SMTP adapter.

Alternatively — resolving the shape at runtime and deleting the frozen attributes removes the class
entirely at the cost of one lookup per send, which is what both the file's own
`# TODO: Remove conditional handling when going 2.0` and the *"Require `gen_smtp` `~> 1.0`, drop
`~> 0.13`"* item in #556 already point at. Your call; the diagnostic above is the cheap half.

### Your reproducible example

_No response_

### Screenshots or Videos

_No response_

### Platform

- OTP: 29 (erts-17.0.2)
- Elixir: 1.20.1
- swoosh 1.27.0, gen_smtp 1.3.0, Linux

`lib/swoosh/adapters/smtp/helpers.ex` is byte-identical at `1.26.3`, `v1.27.0`, `1.27.1` and on
`main` today (fetched and diffed), so this is not fixed in the latest release. The reproducer ran on
1.27.0 only because that was the newest version my Hex registry mirror offered at the time.

### Additional context

Two closed issues share this exact `BadMapError` signature but not this cause, so this is not a
duplicate of either:

- **#578** — gen_smtp 1.1.1 renamed its app file, the `:vsn` probe came back empty, `gen_smtp_major`
  fell to `0`, and users got this stacktrace. Fixed by *hardening the probe* — the
  `Application.load/1` + `Code.ensure_loaded?/1` form in the file today. The silent `else: 0`
  fallback was left in place.
- **#552** — same signature during the gen_smtp 1.0 list→map migration.

Neither covers *gen_smtp genuinely absent when the adapter compiled, present when it runs*, and a
hardened probe cannot help there: there is nothing to probe. That is why this asks for **diagnosis**
rather than reopening version detection — #578 suggests this signature has already cost you more
support time than it should.

Related: **#556** ("Future v2 TODOs") already carries *"Require `gen_smtp` `~> 1.0`, drop
`~> 0.13`"*, which is the v2 form of the alternative fix above.

dajiaohuang:
Fixes #1189.

`Swoosh.Adapters.SMTP.Helpers` fixes the shape of the mimetuple's parameter slots
at compile time, from the `:gen_smtp` that was loaded then. Since `:gen_smtp` is
optional, a build compiled before it was added keeps the gen_smtp 0.x keyword-list
shape while running against gen_smtp 1.x, which reads that slot as a map. Every
send then fails with a `BadMapError` whose frames all point into `:mimemail`, and
nothing in the output says the Swoosh build is stale.

This checks the compiled shape at the single `:mimemail.encode/2` call site: when
the build carries the 0.x shape while gen_smtp 1.x is loaded, it raises an error
naming the cause and the repair (`mix deps.compile swoosh --force`). Builds whose
compile-time and runtime gen_smtp agree are unaffected, and a correctly compiled
build takes the map branch, which costs nothing.

Verified by recompiling Swoosh with `:gen_smtp_client` off the code path and then
running with gen_smtp 1.3.0 present: `Helpers.body/2` raised `BadMapError` before
this change and the new error after it. `mix format --check-formatted` and
`mix test` pass (622 tests, 0 failures).


princemaple:
Hi @dajiaohuang Thanks for the PR. I'm going to close this PR for the [reason I mentioned on the issue comment](https://github.com/swoosh/swoosh/issues/1189#issuecomment-5723803310).

